PR #9, opened 2019-06-19
Backend/#1/mongodb generate connect

ecl1ipse:
Generated a connection the atlas database and made a config locally to get the url.


server/db.ts
```diff
@@ -0,0 +1,29 @@
+//Command to run datase: mongod --dbpath=./data
+
+const MongoClient = require('mongodb').MongoClient
```
Skithy: want to commit the updated await async code instead?

PR #12, opened 2019-06-26
mock api route response

Mysterise:


server/controllers/api.ts
```diff
@@ -0,0 +1,8 @@
+import { Response, Request } from 'express'
+
+/**
+ * GET /api/course/:courseId
+ */
+export let getCourse = (req: Request, res: Response) => {
```
Skithy: export const

Skithy: also is there a way to explicitly define the params in typescript?
something like
```
interface IGetCourseParams {
  courseId: string
}

export const getCourse = (req: Request<IGetCourseParams>, res: Response) => {
```

Mysterise: > export const

fixed

PR #13, opened 2019-07-03
Backend/#10/database api

ecl1ipse:
Created dbCreate, dbRead, dbUpdate and dbDel functions in db_api.ts.


server/db.ts
```diff
@@ -0,0 +1,19 @@
+import { Env, config } from './config'
+const settings = config[process.env.NODE_ENV || Env.DEV]
+import Database from './dbApi'
+
+const test = async () => {
```
Mysterise: Will be moving this stuff out in #16

server/dbApi.ts
```diff
@@ -0,0 +1,70 @@
+import { MongoClient, Db } from 'mongodb'
+import { Env, config } from './config'
+
+const debug = true
+
+// URL for database
+//const env: Env = Env[process.env.NODE_ENV] || Env.development
```
Mysterise: Remove outdated commented line

server/dbApi.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  getDb = async (): Promise<Db> => {
+    if (!this.client) await this.connect()
+    const db = this.client.db(dbName)
+    return db
+  }
+
+  dbCreate = async doc => {
```
Mysterise: @Skithy do we want to be consistent in function parameters, and enforce parenthesis?

server/db.ts
```diff
@@ -0,0 +1,19 @@
+import { Env, config } from './config'
+const settings = config[process.env.NODE_ENV || Env.DEV]
+import Database from './dbApi'
```
Mysterise: shouldn't mix import lines with declarations

server/db.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const test = async () => {
+  console.log('hai')
+  await Database.dbCreate({ id: 'COMP1521', vac: '0', gh: 8 })
+  var doc = await Database.dbRead('COMP1521')
```
Mysterise: Should be const?

server/dbApi.ts
```diff
@@ -0,0 +1,70 @@
...
+
+// URL for database
+//const env: Env = Env[process.env.NODE_ENV] || Env.development
+const url = config[process.env.NODE_ENV || Env.DEV].database
+
+// DB Name
+const dbName = 'Notangles'
+const colName = 'Courses'
+
+// Create a the MongoClient
```
Mysterise: Wrong comment/misplaced? Should be a comment describing the Database class instead.

server/dbApi.ts
```diff
@@ -0,0 +1,70 @@
...
+    var doc = await col.findOne({ id: id })
+    if (debug) console.log('Reading doc: ' + doc)
+    return doc
+  }
+
+  dbUpdate = async (id, doc) => {
+    const db = await this.getDb()
+    const col = db.collection(colName)
+    try {
+      await col.updateOne({ id: id }, { $set: doc })
```
Mysterise: Might be good to have a small comment describing why $set is here for future ref

server/app.ts
```diff
@@ -1,11 +1,13 @@
+import db from './db_api'
+
 const express = require('express')
```
Skithy: convert this one to import as well. try to not mix up imports and requires, keep it consistent

server/config.ts
```diff
@@ -0,0 +1,25 @@
+import * as secret from './secret.json'
```
Skithy: i think, should add a comment here that we should make the secret.json if it you currently don't have one

server/db.ts
```diff
@@ -0,0 +1,19 @@
+import { Env, config } from './config'
+const settings = config[process.env.NODE_ENV || Env.DEV]
+import Database from './dbApi'
+
```
Skithy: have all the imports at the top, then do stuff after
```
import X from ''
import Y from ''

const z = ...
```

server/config.ts
```diff
@@ -0,0 +1,25 @@
...
+  },
+
+  [Env.STAGING]: {
+    database: secret.staging,
+  },
+
+  [Env.PROD]: {
+    database: secret.prod,
+  },
+}
```
Skithy: actually, you can do something like
```
const allConfig: Record<Env, IConfig> = {
  [Env.DEV]: {
    database: secret.dev,
  },

  [Env.STAGING]: {
    database: secret.staging,
  },

  [Env.PROD]: {
    database: secret.prod,
  },
}

export const config = allConfig[process.env.NODE_ENV || Env.DEV]
```
this way, you don't have to check env every time you wanted to use config

server/dbApi.ts
```diff
@@ -0,0 +1,70 @@
+import { MongoClient, Db } from 'mongodb'
+import { Env, config } from './config'
+
+const debug = true
```
Skithy: maybe instead have this in the config instead

server/dbApi.ts
```diff
@@ -0,0 +1,70 @@
...
+    const db = await this.getDb()
+    const col = db.collection(colName)
+    await col.insertOne(doc)
+    if (debug) console.log('Created doc: ' + doc)
+  }
+
+  dbRead = async id => {
+    const db = await this.getDb()
+    const col = db.collection(colName)
+    var doc = await col.findOne({ id: id })
```
Skithy: try not to use var (var actually works differently from let and const, which can lead to some confusing behaviour. if you want, you can research about scopes in javascript and all the quirks of it)
always use const, then let

Skithy: for an object, if the key is the same name as the value variable, you can exclude it
```
col.findOne({ id })
```

server/dbApi.ts
Skithy: maybe dbInsert or dbAdd? dbCreate sounds like you are trying to make a database.

server/dbApi.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  dbCreate = async doc => {
+    const db = await this.getDb()
+    const col = db.collection(colName)
+    await col.insertOne(doc)
+    if (debug) console.log('Created doc: ' + doc)
+  }
+
+  dbRead = async id => {
```
Skithy: i think here, it is good to specify what data-type id should be.
```
dbRead = async (id: string) => {
```
or
```
dbRead = async (id: number) => {
```

server/dbApi.ts
Skithy: same here `col.updateOne({ id }, { $set: doc })`

server/dbApi.ts
Skithy: should be a in-built prettier rule
i think its okay without params for single, looks a bit cleaner
```
const cats = animals.map(animal => animal.type === 'cat')
```
```
const cats = animals.map((animal) => animal.type === 'cat')
```

server/dbApi.ts
Skithy: i think it's better to just refer to the official documentation here, and searching yourself how does updateOne works.

Skithy: Otherwise we end up with tons of comments about third party library stuff.

Mysterise: Agreed, though think this is an exception - most library stuff are fairly intuitive, whereas the $set syntax looks alien with the rest of the codebase so far

PR #19, opened 2019-07-05
Scraper setup

AD9000:
Setup completed with basic puppeteer code that navigates to the timetable page.


scraper/scraper.ts
```diff
@@ -0,0 +1,26 @@
+import * as puppeteer from 'puppeteer'
+const chalk = require("chalk");
```
Skithy: what's chalk?
also if it is a 3rd party library, please include it in package dependencies/devdependencies (npm i -S or npm i -D)
also try to keep the import style consistent. Even though tutorials and example code uses `require(...)`, use `import ... from ...` instead, which works with typescript much better
`import * as chalk from 'chalk'`

scraper/scraper.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const error = chalk.bold.red;
+const success = chalk.keyword("green");
```
Skithy: use `chalk.green` here? it's more typesafe

Skithy: also run prettier over this, since there's a mix of semicolons and no semicolons.
If you vscode, recommend getting vscode plugin and configuring it so it auto-formats after every save, so don't need to worry about it

scraper/scraper.ts
```diff
@@ -0,0 +1,26 @@
...
+
+(async() => {
+    try
+    {
+        // Launch the browser. Headless mode.
+        var browser = await puppeteer.launch();
+        var page = await browser.newPage();
```
Skithy: try to use `const` or `let` where-ever possible, instead of var. var has some very interesting properties, which has the potential for nasty bugs, especially with async operations.
Here's an interesting example. Try to guess what each of them console.logs out, then try running it
```
for (var i = 0; i < 3; i++) console.log(i)
for (let i = 0; i < 3; i++) console.log(i)
for (var i = 0; i < 3; i++) setTimeout(() => console.log(i), 0)
for (let i = 0; i < 3; i++) setTimeout(() => console.log(i), 0)
```

Skithy: Also, would want to add eslint to this project, so it can check these code styling issues automatically for you

scraper/scraper.ts
```diff
@@ -0,0 +1,26 @@
...
+        // Go to the timetable page
+        await page.goto('http://timetable.unsw.edu.au/2019/subjectSearch.html', { waitUntil: 'networkidle2' });
+        
+        // Close the browser.
+        await browser.close();
+        console.log(success("Browser closed"));
+    }
+    catch (err) {
+        console.log(error(err));
+        await browser.close();
```
Skithy: prettty sure you don't have access to the browser variable here, since it was declared within the try block. if you want use a variable that is both used in the try and catch blocks, declare it outside
```
const browser = puppeteer.launch()
try {
  ...
} catch (err) {
  await browser.close()
}
```

Skithy:
nice work! just added some comments

AD9000:
Hi guys, I updated the pull request with a few commits. Added eslint and prettier. 
(Also, Compiler options -> es6 is  for the functions with => arrows.)

scraper/.eslintrc.json
```diff
@@ -0,0 +1,21 @@
+{
+  "env": {
```
Skithy: have to use https://github.com/typescript-eslint/typescript-eslint for this one.
Also some of the rules (indent, linebreak, quotes, semi) should be handled by prettier, not eslint.
```
globals:{
   Atomics, SharedArrayBuffer? 
}
```

scraper/.prettierrc.json
```diff
@@ -0,0 +1,4 @@
+{
```
Skithy: there's already a prettierrc in the parent directory, no need for another one

scraper/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "start": "ts-node scraper.ts",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "prettier-init": "^1.1.0",
```
Skithy: what does this package do?

scraper/.eslintrc.json
AD9000: So, i should install that instead of the current config?

scraper/.prettierrc.json
AD9000: cool. Removed.

scraper/package.json
AD9000: It was supposed to initialize prettier. Didn't work as expected. Must have forgotten to remove it. Removed.

scraper/.eslintrc.json
Skithy: probably better to install it in another pull request. try to only change what is needed

AD9000: Cool. Removed

scraper/scraper.ts
```diff
@@ -0,0 +1,28 @@
+import * as puppeteer from 'puppeteer';
+
+// import * as .. does not work for chalk, so require.
+const chalk = require('chalk');
+
+const error = chalk.bold.red;
+const success = chalk.green;
```
Mysterise: not sure about having console colour modification with the code, may want to remove later but non significant issue.

Issue #20, opened 2019-07-17
Implement exposed express API with internal database API


Mysterise:
Linking the exposed express API in #8 with the internal database API in #10 to create an actually functional API.

Mysterise:
#24

Issue #22, opened 2019-07-24
Frontend fetch course data from backend API


Mysterise:
Fetch data from the API exposed in #20 

1. Fetch list of all courses in the current term from the DB on page load
2. When user selects specific course(s), fetch all classes of that course in the current term

Mysterise:
Merged in #25

PR #24, opened 2019-07-24
Backend/#20/implementing api with db

Mysterise:
#20

### DIFF
--- a/server/controllers/api.ts
+++ b/server/controllers/api.ts
@@ -22,6 +22,6 @@ export const getCourse = async (req: Request, res: Response) => {
       'dbRead returned null course. params: ' +
         util.inspect(params, { showHidden: false, depth: null })
     )
-    res.status(400).send('Invalid courseId param')
+    res.status(400).send('Invalid termId/courseId param')
   }
 }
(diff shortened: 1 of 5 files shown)


server/dbApi.ts
```diff
@@ -27,32 +26,32 @@ class Database {
     return db
   }
 
-  dbAdd = async doc => {
+  dbAdd = async (termColName: string, doc) => {
```
Skithy: dbAdd was meant to be a generic one
maybe make another function that uses dbAdd instead. Can be more specific, add types
```
interface Class {}
addClass = async (termColName: string, class: Class) => {}
```

server/controllers/api.ts
```diff
@@ -1,13 +1,27 @@
...
   const params: IGetCourseParams = req.params
-  res.send(params.courseId)
+  const course = await db.dbRead(params.termId, params.courseId)
+
+  if (course) {
+    res.send(course)
+  } else {
+    console.error(
+      'dbRead returned null course. params: ' +
+        util.inspect(params, { showHidden: false, depth: null })
```
Skithy: hmmm interesting, didn't know this existed

Mysterise: it's very nice!

server/dbApi.ts
Mysterise: will discuss in meeting

Mysterise:
Small related backend change is done in #22

PR #27, opened 2019-08-28
Scraper/#4/scrape one page

AD9000:
Was supposed to scrape one page. My bad I ended up scraping them all...


AD9000:
On a more serious note, there is (at least) two minor issue to solve before I can merge. I have made the scraper into a function so that the database can just call it whenever it wants. But I do not know the exact export style to be used...

AD9000:
Note: instead of ClassUtils, timetable.unsw.edu.au was scraped, it is both easier to scrape and on the unsw official handbook page...

AD9000:
Also, I have got an upto 30 (could be more if needed) pages in parallel scraper working now. Should I add it to this pull request? It runs in 9 minutes and uses around 500 MB of memory.

AD9000:
Scratch my earlier comment. According to my test, the original scraper takes around 80 minutes to run, whereas the latest one runs in about 3 mins. The memory usage is slightly higher for the new one, but should be alright

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
 import * as puppeteer from 'puppeteer';
 
-// import * as .. does not work for chalk, so require.
-const chalk = require('chalk');
+// Remove any html character entities from the given string
+// At this point, it only looks for 3 of them as more are not necessary
```
Mysterise: Good high-level comments for functions, though try to write these in a more docstring style. For example, instead of using "//", use "/**"

```suggestion
/** 
 * Remove any html character entities from the given string
 * At this point, it only looks for 3 of them as more are not necessary
 */
```

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+
+  // &lt --> < (less than), this could be changed to before??
+  newstr = newstr.replace('&lt;', '<');
+
+  // There was no greater than sign found, but if neccessary, can be added here
+
+  return newstr;
+};
+
+// Gets all the class data
```
Mysterise: Same comment as before, use DocString style comments for function descriptions (/**)

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+const getClassData = async (data, rowStartIndex) => {
+  try {
+    // Store data for each class
```
Mysterise: Try to avoid redundant comments like these that don't add much value to our understanding of the code. It's great that you have a lot of emphasis on documentation, however for the more obvious cases like this, comments are more of an annoyance than they help.

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+        }
+      }
+    }
+
     // Close the browser.
     await browser.close();
-    console.log(success('Browser closed'));
+
+    // Return the data that was scraped
+    return JSON.stringify(timetableData);
```
ecl1ipse: The data return doesn't quite match the database format. Would it be possible to return a list of terms? Each term would have courses and each course would only have the classes that run in the respective term. From there, the class is the same as you have it now. This should make the search for classes during a specific term faster.

AD9000: Sounds reasonable. I'll replicate the course object into 1-4 objects depending on how many terms there are.

scraper/scraper.ts
AD9000: alg

scraper/scraper.ts
AD9000: So, I'll keep the docString style comments for the top of each function and single line comments for the rest

scraper/scraper.ts
AD9000: Added.

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+    if (data[notesIndex] !== '' && notesRegex.test(data[notesIndex])) {
+      const result = notesRegex.exec(data[notesIndex]);
+      classData['notes'] = result[1];
+    }
+
+    // Return the scraped class along with updated row index and the term to add
+    // the class to.
+    return [classData, rowStartIndex + 1, term];
+  } catch (err) {
+    throw new Error(err);
```
Skithy: bit useless here... if there's an error, catch the error then throw it again

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+  try {
+    // Store data for each class
+    const classData = {};
+    // term number to append class to
+    let term = 0;
+
+    // First 4 are similar in format and are:
+    // class id, section id, term, activity(lecture/tutorial)
+    let fields = ['classID', 'section', 'term', 'activity'];
+    for (let field of fields) {
```
Skithy: should use `const` when looping using `for ( ... of ...)`

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+      // Store the term number to later append to the correct term
+      if (field === 'term') {
+        if (data[rowStartIndex].charAt(0) === 'U') {
```
Skithy: can probably save `data[rowStartIndex]` to a variable, and use that instead

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+        } else {
+          term = parseInt(data[rowStartIndex].charAt(1));
+        }
+      }
+      if (
+        field === 'activity' &&
+        data[rowStartIndex].includes('Course Enrolment')
+      ) {
+        // Abort if you are looking at course enrolment
+        return [[], rowStartIndex + 9, -1];
```
Skithy: what is `rowStartIndex + 9`?

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+      enrolments: enrAndCap[0],
+      capacity: enrAndCap[1]
+    };
+    rowStartIndex++;
+
+    // Start and end dates, meeting dates, census date -> not needed
+    rowStartIndex += 3;
+
+    // Instruction mode:
+    classData['mode'] = data[rowStartIndex];
```
Skithy: prefer to use the row numbers directly, instead of incrementing a rowStartIndex. Bit hard to keep track of what index you are currently at (e.g. use `data[3]`)

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+
+  return Array.from(urlSet);
+};
+
+const scrapePage = async page => {
+  // Stores data for every course on a page.
+  const coursesData = [];
+  let rowStartIndex = 0;
+  // Get all the data elements.
+  const data = await page.$$eval('.data', element =>
```
Skithy: I think that instead of grabbing all the `.data` items from the page all at once, it's better to go chunk by chunk
Screen Shot 2019-09-12 at 2 32 18 pm
Then you can find the `.data` within this chunk. This way, you don't have to worry about stuff like 'Back to Top' and would be easier to reuse within a function, since all the chunks are very consistent.

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+      )[0].innerHTML;
+      let regexp = /(^[A-Z]{4}[0-9]{4})(.*)/;
+      return regexp.exec(courseHeader);
+    });
+    courseData['courseCode'] = courseHead[1].trim();
+    courseData['name'] = removeHtmlSpecials(courseHead[2].trim());
+
+    // Getting the course type (undergrad/postgrad)
+    while (
+      data[rowStartIndex] !== 'Undergraduate' &&
```
Skithy: i feel like instead of grabbing all the data from the page, and filtering it by keywords, i think it would be easier to just look for the keywords directly.
Instead of looping through all the list of data, looking for 'Undergraduate' keyword, just find the 'Undergraduate' item on the component/page directly.

scraper/scraper.ts
Skithy: It's easier to parse and understand if you split them into individual components, rather than trying to filter out a stream of data

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+    coursesData.push(courseData);
+  }
+  return coursesData;
+};
+
+// Creates browser pages to then use to scrape
+const createPages = async (browser, batchsize) => {
+  // List of pages
+  const pages = [];
+  for (let pageno = 0; pageno < batchsize; pageno++) {
```
Skithy: instead of creating all the pages at the start, then using it for each batch, would it be cleaner to create a page at the start of every batch?

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
     });
 
+    // Defining the regex for course scraping...
+    let regex = /([A-Z]{8})\.html/;
+
+    // Gets all the dataurls on the timetable page.
+    const urlList = await getDataUrls(page, base, regex);
+
+    // Defining the regex for each of the subject codes...
+    regex = /([A-Z]{4}[0-9]{4})\.html/;
```
Skithy: try not to reuse variables. makes it harder to debug, since you have to search for when the last time it has been changed. It's more robust to create a new variable every time, use `const` whenever possible

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+    // Go to the page with list of subjects (Accounting, Computers etc)
+    await page.goto(base, {
       waitUntil: 'networkidle2'
     });
 
+    // Defining the regex for course scraping...
+    let regex = /([A-Z]{8})\.html/;
+
+    // Gets all the dataurls on the timetable page.
+    const urlList = await getDataUrls(page, base, regex);
```
Skithy: could this regex just be inside the function, instead of passing it around?

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+    // array of resolved promises from the promises array
+    let result = [];
+
+    // Jobs -> urls of each subject
+    let jobs = [];
+
+    // Scrape all the urls from the subject pages (eg: COMPKENS, etc)
+    for (let url = 0; url < urlList.length; ) {
+      promises = [];
+      result = [];
```
Skithy: try not to reassign arrays and objects. better off just having the declaration in here
```
const promises = []
const result = []
```

scraper/scraper.ts
```diff
@@ -1,28 +1,480 @@
...
+    let promises = [];
+
+    // array of resolved promises from the promises array
+    let result = [];
+
+    // Jobs -> urls of each subject
+    let jobs = [];
+
+    // Scrape all the urls from the subject pages (eg: COMPKENS, etc)
+    for (let url = 0; url < urlList.length; ) {
```
Skithy: not a fan of non-regular for loops (increments in the loop). Might as well use a while loop instead

scraper/scraper.ts
AD9000: Each class is around 12 lines. 9 is for alignment. I'll use a const for it to make it more readable

scraper/scraper.ts
AD9000: To parse the class, I will have to increment it by 3 anyway. Plus I thought that if we need any of the 3 field that were skipped, we can just parse them there later...

scraper/scraper.ts
AD9000: Sounds like a great idea! I'll see what I can do

scraper/scraper.ts
AD9000: Well, the only labelling the Undergraduate element has is the class data... Plus, if I have to parse the whole data array, I assume it should be more efficient than querying the exact path to the element?

scraper/scraper.ts
AD9000: alg

scraper/scraper.ts
AD9000: The function is called with two different regexes, so no. If the page changes, we can simply change the regex, is the idea

scraper/scraper.ts
AD9000: cool

scraper/scraper.ts
AD9000: Could you tell me what do you mean by 'create a page at the start of each batch'?
Because creating a page just for scraping one link would be very inefficient...

scraper/scraper.ts
AD9000: as in...check the data outside the loop? cool

scraper/scraper.ts
AD9000: all good

AD9000:
I was able to solve both major issues with the scraper:
1. Using chunks for the page scraping makes it much stable (thanks tony dong). We dont have to worry about pages breaking due to small changes now (hopefully lol)
2. Scraper will now return an object with 6 terms ---> [Summer, T1, T2, T3, S1, S2]

Mysterise:
See #30 for most updated scraper progress

PR #46, opened 2020-01-11
Backend/#42/refactoring

ecl1ipse:
Editing the code to adhere to development practices and making the file structure in server neater.


server/src/controllers/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './getCourse'
+export * from './getCourseList'
```
Mysterise: I prefer explicitly importing the named variables and exporting them here, just so you can see exactly what you're getting when you import the index:

```
import { foo } from './wow';
import { bar } from './yeet';

export { foo, bar }
```

Mysterise: And it also allows you to track exactly which var comes from where.

PR #53, opened 2020-01-18
Dockerised server

ecl1ipse:
dockerised the server

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,0 +1,18 @@
+# Grab the latest Node base image
+FROM node:latest
+
+# Set the current working directory inside the container
+WORKDIR /server
+
+# Copy the entire project into the container
+COPY package.json .
+COPY src/secret.json src/secret.json
+
+# Install node modules inside the container using the copied package.json
+RUN npm install
+
+COPY . .
+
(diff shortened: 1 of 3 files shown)


server/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+WORKDIR /server
+
+# Copy the entire project into the container
+COPY package.json .
+COPY src/secret.json src/secret.json
+
+# Install node modules inside the container using the copied package.json
+RUN npm install
+
+COPY . .
```
Skithy: copy itself?

Mysterise: This tells docker to copy from the host directory to the image. The first . indicates the current directory in the host, the second . indicates the current directory in the image.

server/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Grab the latest Node base image
+FROM node:latest
+
+# Set the current working directory inside the container
+WORKDIR /server
+
+# Copy the entire project into the container
+COPY package.json .
+COPY src/secret.json src/secret.json
```
Mysterise: is there a reason why we specifically copy the secret.json here rather than just having that done in line 14?

PR #77, opened 2020-04-21
Feature/37 twelve hour time

lukefisklennon:
Added a button in the top left cell to switch between 24-hour and 12-hour time. (I'm not sure if it looks very good so it's probably a temporary solution until the new nav bar is done). Also added storage of this preference in `localStorage`, and a utility file to make it easier (defaults, storing non-string types).

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -24,7 +24,7 @@ const StyledApp = styled.div`
   padding: 10px 20%;
 
   display: grid;
-  grid-template-rows: 1fr 1fr 90% 
+  grid-template-rows: 1fr 1fr 90%
   grid-template-columns: auto;
 
   text-align: center;
@@ -46,13 +46,16 @@ const App: FunctionComponent = () => {
   const [coursesList, setCoursesList] = useState<CoursesList>([])
   const [selectedCourses, setSelectedCourses] = useState<CourseData[]>([])
   const [selectedClassIds, setSelectedClassIds] = useState<string[]>([])
+  const [twelveHourMode, setTwelveHourMode] = useState<boolean>(true)
+
   const assignedColors = useColorMapper(
     selectedCourses.map(course => course.courseCode)
   )
 
   useEffect(() => {
     fetchClassesList()
-  }, [])
+    window.onkeyup = () => setTwelveHourMode(!twelveHourMode)
+  }, [twelveHourMode])
 
   const handleSelectCourse = async (e: CourseOption) => {
     const selectedCourseClasses = await getCourseInfo('2020', 'T1', e.value)
@@ -126,6 +129,7 @@ const App: FunctionComponent = () => {
             selectedClassIds={selectedClassIds}
             assignedColors={assignedColors}
             onSelectClass={handleSelectClass}
+            twelveHourMode={twelveHourMode}
           />
         </DndProvider>
       </StyledApp>
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -14,6 +14,8 @@ import { getCoursesList } from './api/getCoursesList'
 import { CoursesList } from './interfaces/CourseOverview'
 import { useColorMapper } from './hooks/useColorMapper'
 
+import storage from './utils/storage'
+
 interface CourseOption {
   value: string
   label: string
@@ -46,7 +48,7 @@ const App: FunctionComponent = () => {
   const [coursesList, setCoursesList] = useState<CoursesList>([])
   const [selectedCourses, setSelectedCourses] = useState<CourseData[]>([])
   const [selectedClassIds, setSelectedClassIds] = useState<string[]>([])
-  const [twelveHourMode, setTwelveHourMode] = useState<boolean>(true)
+  const [twelveHourMode, setTwelveHourMode] = useState<boolean>(storage.get('twelveHourMode'))
 
   const assignedColors = useColorMapper(
     selectedCourses.map(course => course.courseCode)
@@ -55,6 +57,10 @@ const App: FunctionComponent = () => {
   useEffect(() => {
     fetchClassesList()
   }, [])
+
+  useEffect(() => {
+    storage.set('twelveHourMode', twelveHourMode)
+  }, [twelveHourMode])
 
   const handleSelectCourse = async (e: CourseOption) => {
     const selectedCourseClasses = await getCourseInfo('2020', 'T1', e.value)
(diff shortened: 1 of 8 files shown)


client/src/components/timetable/TimetableLayout.tsx
```diff
@@ -34,18 +51,46 @@ const TimetableLayout: FunctionComponent<TimetableLayoutProps> = ({
...
+        n -= 12
+        period = 'PM'
+      }
+      return n + ' ' + period
+    } else {
+      let hour = String(n)
+      if (hour.length == 1) {
+        hour = '0' + hour
+      }
+      return hour + ':00'
```
maxwowo, line 91: Not very clean, can be simplified to 
```
return `${String(n).padStart(2, '0')}:00`
```
We can just `padStart` to handle 0 padding, since it pads the string from the front with a given character up to the desired length.

client/src/components/timetable/TimetableLayout.tsx
```diff
@@ -34,18 +51,46 @@ const TimetableLayout: FunctionComponent<TimetableLayoutProps> = ({
...
+    days.map((_, x) =>
       <BaseCell key={x * 1000 + y} x={x + 2} y={y + 2} />)
   )
 
   return (
     <>
-      <BaseCell key={0} x={1} y={1} />
+      <BaseCell key={0} x={1} y={1}>
+        <TwelveHourModeToggle onClick={() => setTwelveHourMode(!twelveHourMode)}>
+          {(twelveHourMode ? '12' : '24') + ' h'}
```
maxwowo, line 62: Change this to 
```
{`${twelveHourMode ? '12' : '24'} h`}
```
for cleaner string formatting

client/src/App.tsx
```diff
@@ -46,6 +48,8 @@ const App: FunctionComponent = () => {
   const [coursesList, setCoursesList] = useState<CoursesList>([])
   const [selectedCourses, setSelectedCourses] = useState<CourseData[]>([])
   const [selectedClassIds, setSelectedClassIds] = useState<string[]>([])
+  const [twelveHourMode, setTwelveHourMode] = useState<boolean>(storage.get('twelveHourMode'))
```
maxwowo, line 51: Change these to `[isTwelveHourMode, setIsTwelveHourMode]` since it's a boolean variable so having `is` in front is more descriptive.

client/src/utils/storage.ts
```diff
@@ -0,0 +1,42 @@
...
+			this.data = JSON.parse(localStorage[storageKey])
+		} else {
+			this.save()
+		}
+	}
+
+	save() {
+		localStorage[storageKey] = JSON.stringify(this.data)
+	}
+}
```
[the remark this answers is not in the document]
lukefisklennon: Fair enough, I've removed the `data` attribute, and have get/set call load each time parse/stringify the JSON. The advantage I see of having it as a class is that all the functions are namespaced together, but I'm happy to change them to standalone functions if you think that'd be better.

client/src/App.tsx
lukefisklennon: I've changed it to `is12HourMode` and `setIs12HourMode` for conciseness.

client/src/constants/defaults.ts
```diff
@@ -0,0 +1,3 @@
+export default {
+	is12HourMode: false
+}
```
maxwowo, line 3: Let's define the type of `defaults` here instead at the location of initialization, that way we can use `defaults` right after importing it, instead of having to define an extra attribute `defaults` in the `storage` class. 

We can just change this to 
```
const defaults: Record<string, any> = {
    is12HourMode: false
}

export default defaults
```

client/src/utils/storage.ts
```diff
@@ -0,0 +1,46 @@
+import defaults from '../constants/defaults'
+
+const STORAGE_KEY = 'data'
+
+const storage = new class {
+	defaults: Record<string, any> = defaults
```
maxwowo, line 6: No longer need this attribute here after defining the type of `defaults` in its own file.

client/src/utils/storage.ts
```diff
@@ -0,0 +1,46 @@
+import defaults from '../constants/defaults'
+
+const STORAGE_KEY = 'data'
+
+const storage = new class {
```
maxwowo, line 5: This class will no longer contain any attributes after removing `defaults`, so there's no reason for it to be a class anymore. Your argument about namespacing still holds though, we can however convert this into an object containing functions instead. It's usages shouldn't need to be changed. But you will need to change all occurrences of `this` to be `storage`.

client/src/utils/storage.ts
```diff
@@ -0,0 +1,46 @@
...
+	defaults: Record<string, any> = defaults
+
+	get(key: string): any {
```
maxwowo, line 8: Change this to `get: (key: string): any => {`, we use arrow functions elsewhere as well so better be consistent ;)

client/src/utils/storage.ts
```diff
@@ -0,0 +1,46 @@
...
+			return data[key]
+		} else if (key in this.defaults) {
+			this.set(key, this.defaults[key])
+			return this.defaults[key]
+		}
+
+		return null
+	}
+
+	set(key: string, value: any) {
```
maxwowo, line 21: Change this to `set: (key: string, value: any) => {`

client/src/utils/storage.ts
```diff
@@ -0,0 +1,46 @@
...
+		let data: Record<string, any> = this.load()
+		data[key] = value
+		this.save(data)
+	}
+
+	load(): Record<string, any> {
```
maxwowo, line 27: Change this to `load: (): Record<string, any> => {`

client/src/utils/storage.ts
```diff
@@ -0,0 +1,46 @@
...
+		if (localStorage[STORAGE_KEY]) {
+			data = JSON.parse(localStorage[STORAGE_KEY])
+		} else {
+			this.save(data)
+		}
+
+		return data
+	}
+
+	save(data: Record<string, any>) {
```
maxwowo, line 39: Change this to `save: (data: Record<string, any>) => {`

client/src/utils/storage.ts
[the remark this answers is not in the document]
maxwowo: That's a good point. There are other methods of namespacing though, one popular approach would be to just have an object containing some functions. I've left a comment in the next review.

client/src/components/timetable/TimetableLayout.tsx
```diff
@@ -34,18 +51,39 @@ const TimetableLayout: FunctionComponent<TimetableLayoutProps> = ({
...
+        </Is12HourModeToggle>
+      </BaseCell>
       {daysCells}
       {hoursCells}
       {otherCells}
     </>
   )
 }
 
+const hourStrings = (range: number[], is12HourMode: boolean): string[] => {
```
maxwowo, line 72: Let's rename this function to `generateHourStrings` instead, since that makes more sense and is more descriptive of what it does. Let's also move it to be above `TimetableLayoutProps` since it's a helper used by `TimetableLayout`.

PR #78, opened 2020-04-21
Frontend/#22/limit course search results

itsViggo:


client/src/App.tsx
```diff
@@ -101,17 +101,37 @@ const App: FunctionComponent = () => {
...
+            options={this.state.options}
+            value={null}
+            onInputChange={this.handleChange}
+            onChange={handleSelectCourse}
+            placeholder="Select a Course"
+          />
+      )
+    }
+
+  }
```
maxwowo, line 127: Let's put the `CourseSelect` component in its own dedicated file, `client/src/components/CourseSelect.tsx` to split up the code. 

This means we need to move line 46, lines 92-97, lines 53-55, as well as the `CourseSelect` component itself to the newly created file.

maxwowo, line 127: Let's use [React Hooks](https://reactjs.org/docs/hooks-intro.html) + function components instead of classes for our components. They're a new addition to React and are quite similar to classes but are defined with a different syntax. Using hooks will help us group the states with their setters together which improves the readability of our code. They're also used elsewhere as well so let's be consistent. 

You can look at other components e.g. `src/components/timetable/TimetableLayouts.tsx` to figure out the syntax of a functional component.

client/src/App.tsx
```diff
@@ -101,17 +101,37 @@ const App: FunctionComponent = () => {
     label: `${course.courseCode} - ${course.name}`,
   }))
 
+  class CourseSelect extends React.Component<{}, {options: Array<CourseOption>}>{
+
+    constructor(props: any){
+      super(props)
+      this.state = {options: courseSelectOptions.slice(0,10)}
+    }
```
maxwowo, line 109: We no longer need this constructor in a function component, we could just do 
```
const [options, setOptions] = React.useState(courseSelectOptions.slice(0,10));
```

client/src/App.tsx
```diff
@@ -101,17 +101,37 @@ const App: FunctionComponent = () => {
...
+
+    handleChange = (inputValue: string, actionMeta: any) => {
+      this.setState({options: courseSelectOptions.filter(x => x.label.toLowerCase().includes(inputValue.toLocaleLowerCase())).slice(0,10)})
```
maxwowo, line 112: Let's define a constant called `NUM_COURSE_OPTIONS`, set it to 10 and use that instead of hard-coding 10 in the code.

maxwowo:
By the way, in the future please use the new branch naming convention! Which is `feature/22-limit-course-search-results`.

PR #79, opened 2020-04-22
Feature/46 new navbar

esyw:
using material UI navbar
image

### DIFF
--- a/client/src/components/Navbar.tsx
+++ b/client/src/components/Navbar.tsx
@@ -1,38 +1,55 @@
-import React from 'react'
+import React from "react";
+import { makeStyles } from "@material-ui/core/styles";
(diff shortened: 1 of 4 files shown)


client/src/components/Navbar.tsx
```diff
@@ -1,38 +1,81 @@
...
+  color: #bde0ff
+`
+const NavButton = styled(Button)`
+  margin-right: 20px;
+`
+
+export default function Navbar() {
+
+  // for dark mode toggle
+  const [selected, setSelected] = React.useState(false);
```
maxwowo, line 50: Let's change these to `isSelected` and `setIsSelected` instead since they're more descriptive

PR #82, opened 2020-04-23
Containerization/32 dockerize client

itsViggo:


lukefisklennon:
Would it be better to change the port we serve on from 3000 to 80 for production? (80 is the default for http servers so you don't need to include the `:80` in the url bar.)

lukefisklennon:
(Or port 443 for https.)

maxwowo:
> 
> 
> Would it be better to change the port we serve on from 3000 to 80 for production? (80 is the default for http servers so you don't need to include the `:80` in the url bar.)

I think by default the built client would run on port 5000. We don't need to worry about having to include the port in the URL because the deployment environment uses a load balancer, which can be configured to forward any request to `notangles.csesoc.unsw.edu.au` to a Docker container instead. It's awesome that you brought it up though, but for our case it doesn't really matter.

client/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Copy the entire project into the container
+COPY . .
+
+# Build server and install npm serve
+RUN npm run build
+RUN npm install -g serve
+
+EXPOSE 3000
+# Run the server
+CMD ["serve","-s","build","-l","3000"]
```
maxwowo, line 22: Let's add an empty line above line 21 to separate the logical sections of the code

maxwowo, line 22: Add a space after each comma in the `CMD` array for better readability

maxwowo:
Also it's a good idea to click "Resolve conversation" for the requested changes that are no longer  relevant so they don't clutter up the PR :)

lukefisklennon:
> > Would it be better to change the port we serve on from 3000 to 80 for production? (80 is the default for http servers so you don't need to include the `:80` in the url bar.)
> 
> We don't need to worry about having to include the port in the URL because the deployment environment uses a load balancer, which can be configured to forward any request to `notangles.csesoc.unsw.edu.au` to a Docker container. It's awesome that you brought it up though, but for our case it doesn't really matter. Port 3000 is a conventional (and also default) port for React apps, so it would make sense to use that instead.

@maxwowo oh I didn't know about that, very cool!

PR #85, opened 2020-04-27
Feature/51 material timetable

lukefisklennon:
This PR adds the material theme primarily to the timetable. Summary of changes:

- Using the material Card class to represent classes.
- Border radius and padding added to classes.
- Changed the style of dropzones to be semi-transparent cards without text. I think this doesn't look as good as it could yet, but making it better might involve changing the dnd dropzone logic which I don't want to do in this PR.
- Changed to material colours for courses (taken from the canonical [material website](https://material.io/resources/color)). All these colours are dark enough so they can all have white text on top of them (for consistency). I'm really bad at choosing colours, so I welcome anyone coming up with a better set.
- Applied the standard material font (Roboto) to the whole page. This PR is meant to just be for the timetable, but it seems like we want everything to use Roboto, and specifying the font for each component individually seems unnecessary.
- Made a central theme file with the colours that are used in multiple places (e.g. primary colour).
- Border radius added to the timetable.
- Changed the timetable borders to be more subtle. The outer-border is now consistent with the inner-borders, and there's no doubling up so everything is 1 px. Also ensuring that it renders fine on retina / high res displays in Chrome, with device pixel ratio detection (this is a quirk in Chrome to do with the sharpness of drop shadows on high res displays).
- I didn't want to touch the inventory much, although I made one change to the font colour for readability, but everything else should be the same.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -23,7 +23,8 @@ export interface CourseOption {
 
 const ContentWrapper = styled.div`
   text-align: center;
-  padding-top: calc(64px + 30px);
+  margin-top: 64px;
+  padding-top: 30px;
   padding-left: 30px;
   padding-right: 30px;
   height: 100vh;
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -23,7 +23,7 @@ export interface CourseOption {
 
 const StyledApp = styled.div`
   height: 85vh;
-  padding: 10px 20%;
+  padding: 10px 15%;
 
   display: grid;
   grid-template-rows: 1fr 1fr 90%
@@ -48,7 +48,7 @@ const App: FunctionComponent = () => {
   const assignedColors = useColorMapper(
     selectedCourses.map(course => course.courseCode)
   )
-  
+
   useEffect(() => {
     storage.set('is12HourMode', is12HourMode)
   }, [is12HourMode])
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -21,12 +21,25 @@ export interface CourseOption {
   label: string
 }
 
-const StyledApp = styled.div`
-  height: 85vh;
-  padding: 10px 15%;
+const ContentWrapper = styled.div`
+  text-align: center;
+  padding-top: calc(64px + 30px);
+  padding-bottom: 30px;
+  padding-left: 30px;
+  padding-right: 30px;
+  height: 100vh;
+  box-sizing: border-box;
+`
+
+const Content = styled.div`
+  width: 1200px;
+  min-width: 600px;
+  max-width: 100%;
+  height: 100%;
+  margin: auto;
 
   display: grid;
-  grid-template-rows: 1fr 1fr 90%
+  grid-template-rows: min-content min-content auto;
   grid-template-columns: auto;
 
   text-align: center;
@@ -91,30 +104,32 @@ const App: FunctionComponent = () => {
   return (
     <div className="App">
       <Navbar />
-      <StyledApp>
-        <SelectWrapper>
-          <CourseSelect
-            onChange={handleSelectCourse}
-          />
-        </SelectWrapper>
-        <DndProvider backend={HTML5Backend}>
-          <Inventory
-            selectedCourses={selectedCourses}
-            selectedClassIds={selectedClassIds}
-            assignedColors={assignedColors}
-            removeCourse={handleRemoveCourse}
-            removeClass={handleRemoveClass}
-          />
-          <Timetable
-            selectedCourses={selectedCourses}
-            selectedClassIds={selectedClassIds}
-            assignedColors={assignedColors}
-            is12HourMode={is12HourMode}
-            setIs12HourMode={setIs12HourMode}
-            onSelectClass={handleSelectClass}
-          />
-        </DndProvider>
-      </StyledApp>
+      <ContentWrapper>
+        <Content>
+          <SelectWrapper>
+            <CourseSelect
+              onChange={handleSelectCourse}
+            />
+          </SelectWrapper>
+          <DndProvider backend={HTML5Backend}>
+            <Inventory
+              selectedCourses={selectedCourses}
+              selectedClassIds={selectedClassIds}
+              assignedColors={assignedColors}
+              removeCourse={handleRemoveCourse}
+              removeClass={handleRemoveClass}
+            />
+            <Timetable
+              selectedCourses={selectedCourses}
+              selectedClassIds={selectedClassIds}
+              assignedColors={assignedColors}
+              is12HourMode={is12HourMode}
+              setIs12HourMode={setIs12HourMode}
+              onSelectClass={handleSelectClass}
+            />
+          </DndProvider>
+        </Content>
+      </ContentWrapper>
     </div>
   )
 }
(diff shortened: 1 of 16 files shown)


lukefisklennon:
Screenshots for comparison:

Old:

image

New:

image

maxwowo:
> 
> 
> Screenshots for comparison:
> 
> Old:
> 
> image
> 
> New:
> 
> image

WOW THIS LOOKS GREAT

lukefisklennon:
> WOW THIS LOOKS GREAT

Hahaha thanks 😆

maxwowo:
I think we should put the course code and activity in one line so that there's more space to fit the location of a class
image

maxwowo:
Let's also make the course code have a bold font to distinguish it from the rest of the card

maxwowo:
Let's make the inventory classes have the rounded corners to be consistent
image

lukefisklennon:
> I think we should put the course code and activity in one line so that there's more space to fit the location of a class

To save space, should we use the class type abbreviation instead of full name (e.g. "LEC" instead of "Lecture")?

lukefisklennon:
> Let's also make the course code have a bold font to distinguish it from the rest of the card

> Let's make the inventory classes have the rounded corners to be consistent

Yup ok

maxwowo:
> > I think we should put the course code and activity in one line so that there's more space to fit the location of a class
> 
> To save space, should we use the class type abbreviation instead of full name (e.g. "LEC" instead of "Lecture")?

Yep I think that's reasonable, I was thinking we should make the timetable wider as well because we want to also display the location of each class

lukefisklennon:
> Yep I think that's reasonable, I was thinking we should make the timetable wider as well because we want to also display the location of each class

Good idea, there's currently lots of unused space either side

maxwowo:
Also I noticed that there's quite a lot of errors in the console
image

lukefisklennon:
Ohhh I didn't think to check the console, I'll have a look

maxwowo:
> > Yep I think that's reasonable, I was thinking we should make the timetable wider as well because we want to also display the location of each class
> 
> Good idea, there's currently lots of unused space either side

Exactly haha, you may have to edit the `StyledApp` component in `App.tsx` to make sure that the course dropdown and inventory have the same width as the timetable

lukefisklennon:
> Exactly haha, you may have to edit the `StyledApp` component in `App.tsx` to make sure that you also change the width of the course dropdown and inventory

Ok cool

Btw in the screenshot you sent, the error seems to be coming from Navbar.tsx which I haven't made any changes to

lukefisklennon:
Although that error isn't showing up in the live version, so it's probably something I've done somehow

maxwowo:
> > Exactly haha, you may have to edit the `StyledApp` component in `App.tsx` to make sure that you also change the width of the course dropdown and inventory
> 
> Ok cool
> 
> Btw in the screenshot you sent, the error seems to be coming from Navbar.tsx which I haven't made any changes to

Oh right, I'll look into it

maxwowo:
> Although that error isn't showing up in the live version, so it's probably something I've done somehow

Oh that's because React does not produce these errors for apps that have been built (with `npm run build`), which is why you're not seeing it in the live version

lukefisklennon:
Ahh ok cool

maxwowo:
Actually there are errors related to the timetable towards the bottom of the console
image

lukefisklennon:
Hmm, does that warning also come up on dev? I can't seem to find that message when I run it, and I don't remember changing anything to do with `isDragging`

maxwowo:
> Hmm, does that warning also come up on dev? I can't seem to find that message when I run it, and I don't remember changing anything to do with `isDragging`

It doesn't seem to come up on dev. That's quite strange, can you try restarting npm and see if that changes anything?

maxwowo:
Oh actually are you using Chrome or Firefox?

lukefisklennon:
Still not showing up after restarting. I'm using Chrome (I don't have Firefox installed so can't test it on there.)

maxwowo:
> Still not showing up after restarting. I'm using Chrome (I don't have Firefox installed so can't test it on there.)

Ah I see, I suspect these warnings might only show up on Firefox, since we realized last year that DND actually behaves differently on the two browsers (it was less clunky on Firefox for some reason). I'll download Chrome now and see if the warnings show up. If not, can you get Firefox and try to fix these issues? These warnings don't appear on dev so it's probable that they were introduced by some changes made in this PR

maxwowo:
Wait, they also show up on Chrome. That's very strange, I'll look into it tomorrow. In the meantime can you make the other changes we talked about? If you're done with those, try to reproduce the warnings on your end
image

lukefisklennon:
The errors have started to appear on my end now. (It was because I needed to select a class for the error to come up.) I think it has something to do with passing custom props to a styled component which extends a material component.

For example, I changed this:
```js
const StyledCell = styled.div<{
  classTime: Period
  ...
}>`
  display: inline-flex;
  ...
`
```
to this:
```js
const StyledCell = styled(Card)<{
  classTime: Period
  ...
}>`
  display: inline-flex;
  ...
`
```
which causes warnings.

I'll keep looking into it, but at this point I don't know why it would be complaining about this.

maxwowo:
> 
> 
> The errors have started to appear on my end now. (It was because I needed to select a class for the error to come up.) I think it has something to do with passing custom props to a styled component which extends a material component.
> 
> For example, I changed this:
> 
> ```js
> const StyledCell = styled.div<{
>   classTime: Period
>   ...
> }>`
>   display: inline-flex;
>   ...
> `
> ```
> 
> to this:
> 
> ```js
> const StyledCell = styled(Card)<{
>   classTime: Period
>   ...
> }>`
>   display: inline-flex;
>   ...
> `
> ```
> 
> which causes warnings.
> 
> I'll keep looking into it, but at this point I don't know why it would be complaining about this.

I see. I suspect the issue may be something to do with Card not expecting a prop called `isDragging`. Can you try keeping it as `styled.div` but make it wrap the `Card` component?

lukefisklennon:
Oh yeah that should work, I'll try it

client/src/App.tsx
```diff
@@ -21,12 +21,24 @@ export interface CourseOption {
   label: string
 }
 
-const StyledApp = styled.div`
-  height: 85vh;
-  padding: 10px 20%;
+const ContentWrapper = styled.div`
+  text-align: center;
+  padding-top: calc(64px + 30px);
```
maxwowo, line 26: Should this just be `padding-top: 94px`

lukefisklennon: Yeah it's a bit hacky, it's meant to be saying offset by 64px for the nav bar and then 30px for the additional padding. I'll split them up into margin and padding properties so it's more readable. (It's like this because the nav bar now has a fixed position when you scroll, which means it doesn't offset other elements the DOM.)

lukefisklennon: Added a commit for it

maxwowo: Ohh I see what you're doing, yeah ok that sounds good

lukefisklennon:
@esyw Thanks! Yeah I'm not completely sure about the rounded corners / gaps either, but it should be pretty easy to change later on if we decide against it

itsViggo:
The mui cards look a lot cleaner than the divs, not sure about the rounded edges but might just be subjective. Overall looks really cool though.

PR #89, opened 2020-04-29
Bugfix/72 client spams course root upon initialization

itsViggo:
- I believe the bug was caused by fetchClassesList being called everytime coursesList was updated and since fetchClassesList updates coursesList, this caused a loop where fetchClassesList was called repeatedly

- The fix was to take coursesList out of useEffect's second argument, leaving at as an empty array so it's only called once

- Also refactored getCourseOptions so it could take in any CoursesList and output a list of CourseOptions


client/src/components/CourseSelect.tsx
```diff
@@ -12,45 +12,47 @@ const StyledSelect = styled(Select)`
...
+      value: course.courseCode,
+      label: `${course.courseCode} - ${course.name}`,
     }))
+  }
 
-    const handleChange = (inputValue: string) => {
-        setOptions(x => courseSelectOptions.filter(x => x.label.toLowerCase().includes(inputValue.toLocaleLowerCase())).slice(0,NUM_COURSE_OPTIONS))
-    }
+  const handleChange = (inputValue: string) => {
+    setOptions(x => courseSelectOptions(coursesList).filter(x => x.label.toLowerCase().includes(inputValue.toLocaleLowerCase())).slice(0, NUM_COURSE_OPTIONS))
```
maxwowo, line 35: It's awesome that you're thinking about refactoring, but I think in this case it's not a good idea to have `courseSelectOptions` as a function, because every time when you type something in the search bar, it would call `courseSelectOptions` which would `map` the coursesList and then `filter` it, which is 2 list traversals compared to when we had `courseSelectOptions` as a constant so only had to use `filter`. Although it doesn't change the theoretical complexity of the code, it does affect the absolute time and that's quite important for something like searching where speed really helps to improve the user experience. We also don't seem to be gaining that much from the refactor as well, so let's change `courseSelectOptions` back to a constant instead.

client/src/components/CourseSelect.tsx
```diff
@@ -12,45 +12,47 @@ const StyledSelect = styled(Select)`
...
+  React.useEffect(() => {
+    fetchClassesList()
+  }, [])
 
-    const courseSelectOptions: CourseOption[] = coursesList.map(course => ({
-        value: course.courseCode,
-        label: `${course.courseCode} - ${course.name}`,
-    }))
+  React.useEffect(() => {
+    setOptions(courseSelectOptions.slice(1, 10))
```
maxwowo, line 27: Shouldn't the range of the slice be `0, 10` since using `1` as the starting index would skip the first course option?

maxwowo: Oh and let's also change `10` to `NUM_COURSE_OPTIONS`, can you also move `NUM_COURSE_OPTIONS` to the top between the last `import` and `StyledSelect` separated by 1 new line above and below?

PR #90, opened 2020-04-29
Feature/36 implement dark mode

itsViggo:
Implemented night mode but also changed react-select to an mui autocomplete and changed divs to mui boxes in order to take advantage of mui's theme provider.

Screen Shot 2020-04-29 at 9 32 48 pm

Screen Shot 2020-04-29 at 9 32 55 pm


lukefisklennon:
Wow, this looks great! 🤩 I really like the new material select as well.

One thing we might want to add as a feature is detecting the system dark mode and automatically applying dark mode if it's enabled, [like this](https://stackoverflow.com/a/57795495/5583289).

Regarding how the preference is stored, I noticed that the default value is missing from `constants/defaults.ts` (I probably should document this better). It seems to work currently, but probably best to add it in there.

There also seems to be a small bug with the new select component, where when you click in the select box in light mode, the text colour changes to white, and when you click it in dark mode, the colour changes to black. Is that issue happening for anyone else?

Another small improvement (perhaps for a future PR) is to make the contrast lower in some places, like the timetable where the lines currently have a very high contrast to the background in dark mode. Passing `isDarkMode` everywhere is probably a hassle, so maybe we should look into [React context](https://reactjs.org/docs/context.html) or something similar?

Other than that it's looking good!

maxwowo:
> 
> 
> could we chuck a `transition: all 0.25s linear;` onto body or smth to not risk the switch being too jarring?
> also any way to keep the select label's contrast? mayb we can just use the outlined variant to not worry abt the shading?
> image

Actually I think we should make the background light grey (select label's background colour) since right now it's a bit too bright

maxwowo:
> 
> 
> > could we chuck a `transition: all 0.25s linear;` onto body or smth to not risk the switch being too jarring?
> > also any way to keep the select label's contrast? mayb we can just use the outlined variant to not worry abt the shading?
> > image
> 
> Actually I think we should make the background light grey (select label's background colour) since right now it's a bit too bright

That'll have to be in a different issue and it's not too important for us right now anyway.

client/src/components/CourseSelect.tsx
```diff
@@ -33,26 +35,46 @@ const CourseSelect: React.FC<CourseSelectProps> = ({ onChange }) => {
...
-      placeholder="Select a Course"
-    />
+    <StyledSelect>
+      <Autocomplete
+        id="combo-box-demo"
+        options={options}
+        getOptionLabel={(option) => option.label}
+        onChange={handleChange}
+        inputValue={inputValue}
+        disableClearable
```
maxwowo, line 65: Let's also add `blurOnSelect` here to make it so that selecting an option causes the autocomplete to lose focus

client/src/App.tsx
```diff
@@ -102,35 +106,49 @@ const App: FunctionComponent = () => {
...
-              onSelectClass={handleSelectClass}
+    <StylesProvider injectFirst>
+      <MuiThemeProvider theme={isDarkMode ? darkTheme : lightTheme}>
+        <ThemeProvider theme={isDarkMode ? darkTheme : lightTheme}>
+          <div className="App">
+            <Helmet>
+              <style>
+                {isDarkMode ? 'body { background-color: #202020; }' : 'body { background-color: white; }'}
+              </style>
+            </Helmet>
```
maxwowo, line 117: This seems quite hacky, can we instead make a `StyledApp` component and make its `background-color` use the current theme's main color, like `TimetableLayout` line 10? We can then replace line 112 with that `StyledApp` component to achieve the same effect

client/src/App.tsx
```diff
@@ -43,28 +47,28 @@ const Content = styled.div`
...
+  const [isDarkMode, setIsDarkMode] = useState<boolean>(storage.get('isDarkMode'))
 
   const assignedColors = useColorMapper(
     selectedCourses.map(course => course.courseCode)
   )
 
   useEffect(() => {
     storage.set('is12HourMode', is12HourMode)
-  }, [is12HourMode])
+    storage.set('isDarkMode', isDarkMode)
```
maxwowo, line 70: Let's put this in its own `useEffect` so that the other option doesn't have to be unnecessarily set when one of them changes

client/src/App.tsx
```diff
@@ -10,18 +11,21 @@ import { CourseData } from './interfaces/CourseData'
...
 
 import { getCourseInfo } from './api/getCourseInfo'
-import { CoursesList } from './interfaces/CourseOverview'
 import { useColorMapper } from './hooks/useColorMapper'
-
+import { Helmet } from 'react-helmet'
 
 import storage from './utils/storage'
+import Box from '@material-ui/core/Box'
+import { MuiThemeProvider } from '@material-ui/core'
```
maxwowo, line 19: Change this to `import { MuiThemeProvider, Box } from '@material-ui/core'` instead

client/src/App.tsx
```diff
@@ -10,18 +11,21 @@ import { CourseData } from './interfaces/CourseData'
 import CourseSelect from './components/CourseSelect'
 
 import { getCourseInfo } from './api/getCourseInfo'
-import { CoursesList } from './interfaces/CourseOverview'
 import { useColorMapper } from './hooks/useColorMapper'
-
+import { Helmet } from 'react-helmet'
 
 import storage from './utils/storage'
+import Box from '@material-ui/core/Box'
```
maxwowo, line 18: We don't need this line

client/package.json
```diff
@@ -6,17 +6,21 @@
...
     "@material-ui/lab": "^4.0.0-alpha.50",
+    "@types/helmet": "0.0.46",
     "@types/jest": "24.0.13",
     "@types/node": "12.0.8",
     "@types/react": "16.8.19",
     "@types/react-beautiful-dnd": "^11.0.2",
     "@types/react-dom": "16.8.4",
+    "@types/react-helmet": "^5.0.15",
     "@types/react-select": "^2.0.19",
+    "darkmode-js": "^1.5.5",
```
maxwowo, line 17: Do we use this library anywhere?

maxwowo:
> 
> 
> Wow, this looks great! 🤩 I really like the new material select as well.
> 
> One thing we might want to add as a feature is detecting the system dark mode and automatically applying dark mode if it's enabled, [like this](https://stackoverflow.com/a/57795495/5583289).
> 
> Regarding how the preference is stored, I noticed that the default value is missing from `constants/defaults.ts` (I probably should document this better). It seems to work currently, but probably best to add it in there.
> 
> There also seems to be a small bug with the new select component, where when you click in the select box in light mode, the text colour changes to white, and when you click it in dark mode, the colour changes to black. Is that issue happening for anyone else?
> 
> Another small improvement (perhaps for a future PR) is to make the contrast lower in some places, like the timetable where the lines currently have a very high contrast to the background in dark mode. Passing `isDarkMode` everywhere is probably a hassle, so maybe we should look into [React context](https://reactjs.org/docs/context.html) or something similar?
> 
> Other than that it's looking good!

Oh wow that post about detecting system dark mode is very interesting, let's definitely do that in the future! I'll make an issue about it in the backlog, it won't be of the highest priority right now though since there's more important stuff to be done. 

Regarding React context, I'm thinking about making the next dev sesh about discussing UI approaches as well as a refactoring session, where we can try to make our code better and more maintainable, a big part of it will be refactoring to use React context or equivalent for 12H time and dark mode. What do you guys think about this? We could also make this a tradition where we would do a major refactor at the start of every term to ensure a good foundation to build upon for the rest of the term :)

lukefisklennon:
> Regarding React context, I'm thinking about making the next dev sesh about discussing UI approaches as well as a refactoring session, where we can try to make our code better and more maintainable, a big part of it will be refactoring to use React context or equivalent for 12H time and dark mode. What do you guys think about this? We could also make this a tradition where we would do a major refactor at the start of every term to ensure a good foundation to build upon for the rest of the term :)

Sounds like a good plan 👍

client/src/App.tsx
maxwowo: Right, I think we can fix this by removing line 34 in `App.tsx`, since that gives the content a fixed height which causes the page to extend outside the root component.

client/src/components/CourseSelect.tsx
maxwowo: Are you still going to address this comment

client/src/components/timetable/TimetableLayout.tsx
```diff
@@ -1,13 +1,13 @@
 import React, { FunctionComponent } from 'react'
 import styled from 'styled-components'
-import * as theme from '../../constants/theme'
+import Box from '@material-ui/core/Box'
```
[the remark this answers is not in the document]
maxwowo: Are you still going to address this comment

PR #92, opened 2020-05-01
Bugfix/83 dark background to end

lukefisklennon:
Fixes an issue where the background in dark mode doesn't extend to the end of the page on large / zoomed out screens:

### DIFF
--- a/client/src/index.css
+++ b/client/src/index.css
@@ -9,10 +9,6 @@ body {
   -moz-osx-font-smoothing: grayscale;
 }
 
-#root {
-  height: 100%;
-}
-
 code {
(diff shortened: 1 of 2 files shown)


client/src/index.css
```diff
@@ -1,10 +1,18 @@
...
   margin: 0;
   font-family: 'Roboto', sans-serif;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
 }
 
+#root {
+  height: 100%;
+}
+
```
maxwowo, line 15: Let's instead make a component
```
const StyledRoot = styled(App)`
  height: 100%;
`
```
in `index.tsx` to use styled components as much as possible.

PR #93, opened 2020-05-01
Feature/80 fix client lint errors

maxwowo:
This PR fixes all of the lint errors in the current code reported by ESLint.

### DIFF
--- a/client/.eslintignore
+++ b/client/.eslintignore
@@ -6,3 +6,7 @@ build
 
 # Don't lint nyc coverage output
 coverage
+
+# Don't lint default files
+serviceWorker.ts
+react-app-env.d.ts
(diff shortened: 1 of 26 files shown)


lukefisklennon:
Looks good. I noticed it inserted semicolons everywhere, is that the new style we should follow?

maxwowo:
> Looks good. I noticed it inserted semicolons everywhere, is that the new style we should follow?

Yep please follow this style! With that said, you don't actually need to write semicolons in your code, since running `npm run lint:fix` will do that for you (as well as convert double quotes to single quotes, I recall we were talking about that a couple weeks ago ;)) So it's a good idea to just run the linter before you commit / push. In the future it'll be automated to make it even more convenient

lukefisklennon:
Neat, I'll do that for my next commit 👍

PR #96, opened 2020-05-01
bugfix/81-client-allows-duplicated-course-selections

itsViggo:
Fixed the bug by checking if the course code is already in the list of selected courses before adding it to the list of selected courses.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -73,10 +73,11 @@ const App: FunctionComponent = () => {
   }, [isDarkMode]);
 
   const handleSelectCourse = async (e: CourseOption) => {
-    const selectedCourseClasses = await getCourseInfo('2020', 'T2', e.value);
-
-    if (selectedCourseClasses) {
-      setSelectedCourses([...selectedCourses, selectedCourseClasses]);
+    if (selectedCourses.filter(course => course.courseCode == e.value).length == 0) {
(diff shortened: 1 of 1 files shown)


client/src/App.tsx
```diff
@@ -73,10 +73,11 @@ const App: FunctionComponent = () => {
   }, [isDarkMode]);
 
   const handleSelectCourse = async (e: CourseOption) => {
-    const selectedCourseClasses = await getCourseInfo('2020', 'T2', e.value);
-
-    if (selectedCourseClasses) {
-      setSelectedCourses([...selectedCourses, selectedCourseClasses]);
+    if (selectedCourses.filter(course => course.courseCode == e.value).length == 0) {
```
maxwowo, line 76: Let's change this line to `if (selectedCourses.find(course => course.courseCode === e.value) === undefined)`, since that seems a bit more straight forward than filtering the list and then checking its length

maxwowo: Also let's not use `==` in the code since its behavior can be [unexpected at times](https://stackoverflow.com/a/9751114)

Don't forget to use the linter by running `npm run lint`, it would've picked up on this and notified you about it.

maxwowo:
This PR's title is a bit vague as well, it's alg to just use the branch name, I think that's the PR's title by default as well when you first create it

PR #98, opened 2020-05-17
Feature/56 about modal

esyw:


lukefisklennon:
Dark theme toggle seems to be broken for me, is that happening for anyone else?

maxwowo:
> 
> 
> Dark theme toggle seems to be broken for me, is that happening for anyone else?

Hmm odd, it works for me

lukefisklennon:
Never mind I had an outdated version of the branch

client/src/components/About.tsx
```diff
@@ -0,0 +1,95 @@
...
+  position: absolute;
+  right: 10px;
+  top: 10px;
+`;
+const DialogContent = styled(MuiDialogContent)`
+  padding: 20px;
+`;
+
+const About: React.FC = () => {
+  const [open, setOpen] = React.useState(false);
```
maxwowo, line 28: Let's name the variables `isOpen` and `setIsOpen` instead, since I think `open` is a built-in function in Node so we shouldn't shadow it

client/src/components/About.tsx
```diff
@@ -0,0 +1,95 @@
...
+            Notangles: No more timetable Tangles
+          </Typography>
+          {handleClose ? (
+            <CloseButton
+              aria-label="close"
+              onClick={handleClose}
+            >
+              <CloseIcon />
+            </CloseButton>
+          ) : null}
```
maxwowo, line 62: Do we need to check for `handleClose` here? I think it'll still work if we get rid of the ternary if statement

client/src/components/About.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const About: React.FC = () => {
+  const [open, setOpen] = React.useState(false);
+
+  const handleClickOpen = () => {
+    setOpen(true);
+  };
+  const handleClose = () => {
+    setOpen(false);
+  };
```
maxwowo, line 35: We can merge these two functions into one 
```
const toggleIsOpen = () => {
  setIsOpen(!isOpen);
}
```
and replace both occurrences with this instead

esyw:
> Minor nitpick - can we have some padding at the bottom of the modal? Everything else looks great :)

about how much extra were you thinking?

maxwowo:
> > Minor nitpick - can we have some padding at the bottom of the modal? Everything else looks great :)
> 
> about how much extra were you thinking?

Actually I don't think that 20px padding in `DialogContent` is being applied to the bottom of the component

esyw:
> > > Minor nitpick - can we have some padding at the bottom of the modal? Everything else looks great :)
> > 
> > 
> > about how much extra were you thinking?
> 
> Actually I don't think that 20px padding in `DialogContent` is being applied to the bottom of the component. Let's just give it the same padding as the sides and top

image
 this is what it looked like for me, seems ~20pxish? tho a bit needs to be scrolled to on smaller screens

maxwowo:
> > > > Minor nitpick - can we have some padding at the bottom of the modal? Everything else looks great :)
> > > 
> > > 
> > > about how much extra were you thinking?
> > 
> > 
> > Actually I don't think that 20px padding in `DialogContent` is being applied to the bottom of the component. Let's just give it the same padding as the sides and top
> 
> image
> this is what it looked like for me, seems ~20pxish? tho a bit needs to be scrolled to on smaller screens

Oh ok, that's a bit odd, the bottom 20px isn't really showing up on my end
image
Hmm, that's fine then maybe my laptop is just weird

PR #100, opened 2020-05-19
bugfix/86-class-duplication

itsViggo:
Turns out that classes aren't actually duplicated but they seem like they are because weeks aren't displayed. So to fix this issue, weeks have been added to class cards.

### DIFF
--- a/client/src/components/timetable/DroppedClasses.tsx
+++ b/client/src/components/timetable/DroppedClasses.tsx
@@ -80,7 +80,7 @@ const DroppedClass: FunctionComponent<DroppedClassProps> = ({
       </p>
       <p>{`${classTime.locationShort}`}</p>
       <p>
-        {`${((classTime.time.weeks.includes(",") || classTime.time.weeks.includes("-")) ? "Weeks " : "Week ") + classTime.time.weeks}`}
+        {`${((classTime.time.weeks.includes(',') || classTime.time.weeks.includes('-')) ? 'Weeks ' : 'Week ') + classTime.time.weeks}`}
       </p>
     </StyledCourseClass>
   );
--- a/client/src/components/timetable/DroppedClasses.tsx
+++ b/client/src/components/timetable/DroppedClasses.tsx
@@ -64,6 +64,10 @@ const DroppedClass: FunctionComponent<DroppedClassProps> = ({
     }),
   });
 
+  const isMultipleWeeks = (weeks: string) => (
+    weeks.includes(',') || weeks.includes('-')
+  );
+
   return (
     <StyledCourseClass
       ref={drag}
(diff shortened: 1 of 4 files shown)


client/src/components/timetable/DroppedClasses.tsx
```diff
@@ -79,6 +79,9 @@ const DroppedClass: FunctionComponent<DroppedClassProps> = ({
         </b>
       </p>
       <p>{`${classTime.locationShort}`}</p>
+      <p>
+        {`${((classTime.time.weeks.includes(",") || classTime.time.weeks.includes("-")) ? "Weeks " : "Week ") + classTime.time.weeks}`}
```
maxwowo: Let's put `classTime.time.weeks.includes(",") || classTime.time.weeks.includes("-")` into a function within the same component to improve readability, something like 
```
const isMultipleWeeks = (weeks) => (
    weeks.includes(",") || weeks.includes("-");
)
```

maxwowo: Let's also add a space after each comma in `classTime.time.weeks` so the string looks nicer. We can just use Regex for this by doing `classTime.time.weeks.replace(/,/g, ', ')`

maxwowo: We can also use JS template literals to make the code look a bit nicer by writing `` `${isMultipleWeeks(classTime.time.weeks) ? 'Weeks' : 'Week'} ${classTime.time.weeks.replace(/,/g, ', ')}` `` instead

client/src/components/timetable/DroppedClasses.tsx
```diff
@@ -79,6 +83,12 @@ const DroppedClass: FunctionComponent<DroppedClassProps> = ({
         </b>
       </p>
       <p>{`${classTime.locationShort}`}</p>
+      <p>
+        {`
+          ${isMultipleWeeks ? 'Weeks ' : 'Week '}
```
maxwowo, line 88: Did you mean to call the function

client/src/components/timetable/DroppedClasses.tsx
```diff
@@ -79,6 +83,12 @@ const DroppedClass: FunctionComponent<DroppedClassProps> = ({
...
+          ${classTime.time.weeks.replace(/,/g, ', ')}
```
maxwowo, line 89: Since we're using `classTime.time.weeks` multiple times let's just make a variable for it and use that instead

client/src/components/timetable/DroppedClasses.tsx
```diff
@@ -79,6 +83,12 @@ const DroppedClass: FunctionComponent<DroppedClassProps> = ({
...
+        `}
```
maxwowo, line 90: No need to spread it across multiple lines tbh, it's not too long

client/src/components/timetable/DroppedClasses.tsx
maxwowo, line 88: Instead of putting a space after both "Weeks" and "Week", we can just have a space after the ${...} and it'll have the same effect

PR #102, opened 2020-05-20
Add footer

esyw:


client/src/App.tsx
```diff
@@ -140,6 +147,20 @@ const App: FunctionComponent = () => {
                     onSelectClass={handleSelectClass}
                   />
                 </DndProvider>
+                <Footer>
+                    DISCLAIMER: Notangles is not an official part of the UNSW
+                    enrolment system, and it does not guarantee data accuracy
+                    & optimality.
```
lukefisklennon, line 153: Should we have this match the disclaimer in the about modal? ("While we try our best, Notangles is not an official UNSW site, and cannot guarantee data accuracy or reliability.")

PR #105, opened 2020-05-27
Feature/48 autotimetabling component

esyw:
component to set autotimetabling options with

make the dropdown button not disabled to see options


maxwowo:
image
Can't seem to run the app, it looks like it's trying to find a file in assets which doesn't exist, perhaps you forgot to stage & commit it?

maxwowo:
Would it be possible to make it so that the autotimetabling components are on a separate row below the course select component for small viewport sizes?

lukefisklennon:
What if the "Go" button was inside the drop-down modal? You only need to use it after making a change to the options, so you'll be in the modal anyway. For small viewport sizes (e.g. phones), maybe we could shorten "Auto-timetable options" down to something like "Auto 🡣" so that it can fit inline with the course select bar? Or perhaps shorten it for every screen size, for consistency.

maxwowo:
> 
> 
> What if the "Go" button was inside the drop-down modal? You only need to use it after making a change to the options, so you'll be in the modal anyway. For small viewport sizes (e.g. phones), maybe we could shorten "Auto-timetable options" down to something like "Auto 🡣" so that it can fit inline with the course select bar? Or perhaps shorten it for every screen size, for consistency.

Good point actually, if the button was inside the dropdown menu then there's no need to put it on a separate row

maxwowo:
Would be great if the dropdown menu had the same width as the dropdown button as well, sounds like it could be difficult to do depending on how it's styled tho
image

itsViggo:
I think the differences in style between the course select bar and the auto-timetable options make the UI look a bit weird, especially since the two components are right next to each other. E.g. the course select and the options have different drop down arrows, different color text, one is in all caps and the other isn't. Is it possible to make the two components more consistent?

maxwowo:
> 
> 
> I think the differences in style between the course select bar and the auto-timetable options make the UI look a bit weird, especially since the two components are right next to each other. E.g. the course select and the options have different drop down arrows, different color text, one is in all caps and the other isn't. Is it possible to make the two components more consistent?

Great point. @esyw seems like you're using the `<Button/>` and `<Popover/>` components to simulate the dropdown, but there's actually a [MUI component](https://material-ui.com/components/selects/) made specifically for that. Using the library components will make your life a bit easier and also improves the consistency across components. Can you make the auto-timetabling options use the `<Select/>` component instead?

lukefisklennon:
> Can you make the auto-timetabling options use the `<Select/>` component instead?

I'm guessing it might be difficult to use `<Select/>` in this case, due to the custom layout (e.g. how would you put a "Go" button in a `<Select/>`?) and custom binary checkbox behaviour. If we add drag-and-drop later, it might not be possible to implement in `<Select/>`.

esyw:
> > Can you make the auto-timetabling options use the `<Select/>` component instead?
> 
> I'm guessing it might be difficult to use `<Select/>` in this case, due to the custom layout (e.g. how would you put a "Go" button in a `<Select/>`?) and custom binary checkbox behaviour. If we add drag-and-drop later, it might not be possible to implement in `<Select/>`.

yeah pretty much this, i think i felt just using the popover would give more flexibility to try different layouts for the selections as we were considering having multiple options listed horizontally, sliders etc. also, using the button seemed the easiest way to have the selected options not show in the button itself like the jira dropdown example. i actually used the select component in my first sandbox prototype thingy so could just use that if that works better? otherwise if it still looks bad after trying to match the styling to the select perhaps we could make it deliberately look a bit more different since it serves a slightly different function?

esyw:
> Would be great if the dropdown menu had the same width as the dropdown button as well, sounds like it could be difficult to do depending on how Material UI styles it though
> image

also still figuring this one out, though didn't consider it originally as i was testing with an option description that would be longer than the button for whatever reason 😅

maxwowo:
> > > Can you make the auto-timetabling options use the `<Select/>` component instead?
> > 
> > 
> > I'm guessing it might be difficult to use `<Select/>` in this case, due to the custom layout (e.g. how would you put a "Go" button in a `<Select/>`?) and custom binary checkbox behaviour. If we add drag-and-drop later, it might not be possible to implement in `<Select/>`.
> 
> yeah pretty much this, i think i felt just using the popover would give more flexibility to try different layouts for the selections as we were considering having multiple options listed horizontally, sliders etc. also, using the button seemed the easiest way to have the selected options not show in the button itself like the jira dropdown example. i actually used the select component in my first sandbox prototype thingy so could just use that if that works better? otherwise if it still looks bad after trying to match the styling to the select perhaps we could make it deliberately look a bit more different since it serves a slightly different function?

Yep that makes sense, let's keep it that way

maxwowo:
> > Would be great if the dropdown menu had the same width as the dropdown button as well, sounds like it could be difficult to do depending on how Material UI styles it though
> > image
> 
> also still figuring this one out, though didn't consider it originally as i was testing with an option description that would be longer than the button for whatever reason 😅

That's alright, don't sweat too much about this since it's just a nitpick. Just leave it if it can't be done without some hacks

maxwowo:
What do you guys think about making the autotimetable dropdown use the outlined variant and the csesoc blue instead of grey filled as it is now? 
image

client/src/components/Autotimetabler.tsx
```diff
@@ -0,0 +1,190 @@
...
+                exclusive
+                value={optionState}
+                onChange={handleOptionChange}
+                aria-label="option choices"
+              >
+                <StyledOptionButtonToggle value="off" aria-label="default">
+                  off
+                </StyledOptionButtonToggle>
+                {optionChoices.map((op) => (
+                  <>
```
maxwowo, line 77: I tried to comment out the `disabled` in `line 119` to check out what the popover looked like, but it didn't render the option buttons correctly, could it be because these `<>` tags are preventing the correct styles from being applied to the option buttons? 
image

client/src/components/Autotimetabler.tsx
```diff
@@ -0,0 +1,190 @@
...
+    }
+  };
+
+  return (
+    <div>
+      <ListItem>
+        <ListItemText
+          primary={optionName}
+          secondary={(
+            <>
```
maxwowo, line 65: Do we need a `<>` tag here? Should be alright without it

client/src/components/Autotimetabler.tsx
```diff
@@ -0,0 +1,190 @@
...
+        anchorOrigin={{
+          vertical: 'bottom',
+          horizontal: 'right',
+        }}
+        transformOrigin={{
+          vertical: 'top',
+          horizontal: 'right',
+        }}
+      >
+        <List dense>
```
maxwowo, line 143: What do you think about making the list not dense? Right now it's looking a little bit crammed, it might be better to just leave the list not dense since there's plenty of space in the browser window anyway and conditionally turn on dense if on mobile (which we'll get to in the future)

client/src/components/Autotimetabler.tsx
```diff
@@ -0,0 +1,190 @@
...
+            optionName="Breaks between classes"
+            optionState={breaksBetweenClasses}
+            setOptionState={setBreaksBetweenClasses}
+            optionChoices={['least', 'most']}
+          />
+
+        </List>
+        <ExecuteButton variant="contained" color="primary" disableElevation onClick={handleClose}>
+          GO
+          <FlashOnIcon />
```
maxwowo, line 184: What do you think about putting the flash icon in front of the "GO"? Since in the case of action buttons, if the icon comes before the text, then you can associate the action before reading the text, so you don't even need to read it most of the times

client/src/components/Autotimetabler.tsx
esyw: aa forgot to get rid of those after removing smth, my bad

client/src/components/Autotimetabler.tsx
esyw: these ones i don't remember putting in, might've been the linter?

maxwowo: Hmm maybe

client/src/components/Autotimetabler.tsx
```diff
@@ -0,0 +1,186 @@
...
+      </ListItem>
+    </div>
+
+  );
+};
+
+interface AutotimetablerProps {
+  isDarkMode: boolean
+}
+const Autotimetable: React.FC<AutotimetablerProps> = ({ isDarkMode }) => {
```
maxwowo, line 92: Should this be called `Autotimetabler` instead?

PR #106, opened 2020-05-27
Feature/135 show class capacity info on cards

itsViggo:
class capacity info now gets shown on cards
Screen Shot 2020-05-27 at 7 49 08 pm

### DIFF
--- a/client/src/components/timetable/DroppedClasses.tsx
+++ b/client/src/components/timetable/DroppedClasses.tsx
@@ -91,12 +91,7 @@ const DroppedClass: FunctionComponent<DroppedClassProps> = ({
       </p>
       <p>{`${classTime.locationShort}`}</p>
       <p>{`${isMultipleWeeks(weeks) ? 'Weeks' : 'Week'} ${weeks.replace(/,/g, ', ')}`}</p>
-      <p>
-        {enrolments}
-        /
-        {capacity}
-        enrolled
-      </p>
+      <p>{`${enrolments}/${capacity} enrolled`}</p>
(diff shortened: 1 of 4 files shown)


client/src/components/timetable/DroppedClasses.tsx
```diff
@@ -49,13 +50,17 @@ interface DroppedClassProps {
   courseCode: string
   color: string
   classTime: Period
+  enrolments: number
```
maxwowo, line 53: Does your code editor complain that enrolments is not a word? Apparently it should be spelt as enrollments instead ¯\_(ツ)_/¯

client/src/components/timetable/DroppedClasses.tsx
```diff
@@ -86,6 +91,7 @@ const DroppedClass: FunctionComponent<DroppedClassProps> = ({
       </p>
       <p>{`${classTime.locationShort}`}</p>
       <p>{`${isMultipleWeeks(weeks) ? 'Weeks' : 'Week'} ${weeks.replace(/,/g, ', ')}`}</p>
+      <p>{`${enrolments}/${capacity} enrolled`}</p>
```
maxwowo, line 94: What do you think about switching the ordering of weeks and enrollment? Since I think enrollment data is more important than the class weeks because it matters more to the users at the timetable planning stage, so it makes more sense to be above class weeks

itsViggo: yep that makes sense to me, i'll switch the order around

PR #109, opened 2020-06-08
Feature/149 auto populate timetable

lukefisklennon:
- Naive timetable auto-population by selecting the first available classes (until auto-timetabling is done)
- Is in preparation for moving inventory to the side of the timetable, becoming an optional space to move classes out of the way to
- Hides inventory so that you can better see what the layout will look like (everything should still work bar moving classes out of the way)

### DIFF
--- a/client/src/components/timetable/Timetable.tsx
+++ b/client/src/components/timetable/Timetable.tsx
@@ -11,8 +11,9 @@ const rows: number = hoursRange[1] - hoursRange[0] + 1;
 
 const StyledTimetable = styled(Box)`
   display: grid;
-  min-height: 700px;
+  min-height: 900px;
   max-height: 900px;
+  margin-top: 15px;
   margin-bottom: 20px;
   box-sizing: content-box;
   border-radius: ${(props) => props.theme.shape.borderRadius}px;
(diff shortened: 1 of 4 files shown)


itsViggo:
This is nice. Makes the UI a lot cleaner by not having the inventory. The auto populate is a bit slow though.

client/src/components/timetable/Timetable.tsx
```diff
@@ -7,16 +7,16 @@ import TimetableLayout from './TimetableLayout';
 import ClassDropzones from './ClassDropzones';
 import DroppedClasses from './DroppedClasses';
 
-const defaultMinTableHeight = 700;
-const defaultMaxTableHeight = 900;
+const rowHeight = 100;
```
maxwowo, line 10: Is there a benefit to using the row's height instead of the timetable's height? Since when we think about changing the height of the timetable it's a bit easier to visualize and change the height of the timetable component as a whole rather than adjusting the height of each individual cell. What do you think?

lukefisklennon: I changed it to row height because I thought it'd be a bit simpler and clearer than what we had before:
```tsx
min-height: ${(props) => props.rows * (defaultMinTableHeight / (defaultEndTime - defaultStartTime))}px;
```
What it is now:
```tsx
min-height: ${(props) => props.rows * rowHeight}px;
```
I'm happy to revert back or change it to something different, what do you think?

maxwowo: Hmm ok, that's fair let's just leave it as it is now

PR #111, opened 2020-06-09
add some colours from material ui palette

esyw:


lukefisklennon:
Should the primary app colour be included in these course colours (i.e. like now, where the first colour chosen is the primary blue colour)? Otherwise looks good.

maxwowo:
> Should the primary app colour be included in these course colours (i.e. like now, where the first colour chosen is the primary blue colour)? Otherwise looks good.

Hmm I reckon it's better if the primary color is not used in the classes since that'll make the classes stand out from the rest of the interface, which is a good thing otherwise it might look a bit bland

lukefisklennon:
> > Should the primary app colour be included in these course colours (i.e. like now, where the first colour chosen is the primary blue colour)? Otherwise looks good.
> 
> Hmm I reckon it's better if the primary color is not used in the classes since that'll make the classes stand out from the rest of the interface, which is a good thing, otherwise it might look a bit bland

Fair enough

PR #113, opened 2020-06-13
feature/154 show more info on class card hover

itsViggo:
Removed all info except for course name from cards until the card is hovered over. On hover, font size is reduced and more info on the class is revealed.

### DIFF
--- a/client/src/components/timetable/DroppedClasses.tsx
+++ b/client/src/components/timetable/DroppedClasses.tsx
@@ -35,13 +35,30 @@ const StyledCourseClass = styled(Card).withConfig({
   margin: 2px;
   position: relative;
   bottom: 0.5px;
+  span {
+    opacity : 0;
+    transition: opacity 0.5s;
+  };
 
   p {
     margin: 0 0;
+    position: relative;
+    top: 1.8rem;
     width: 100%;
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
+    transition: all 0.5s;
+  }
+
+  &:hover {
+    span {
+      opacity: 1;
+    };
+    p {
+      top: 0rem;
+      font-size: 0.7rem;
+    };
   }
 `;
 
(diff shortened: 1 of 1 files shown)


lukefisklennon:
Hey, this looks really good! A small improvement could be to make it feel a bit more snappy, by lowering the transition time from 0.5s to something like 0.1-0.2s. What do you reckon?

lukefisklennon:
Also what do you think about hiding the course name altogether on hover? Would this be too confusing for users?

maxwowo:
> Hey, this looks really good! A small improvement could be to make it feel a bit more snappy, by lowering the transition time from 0.5s to something like 0.1-0.2s. What do you reckon?

Good point, 0.1-0.2s feels a bit too quick though, I'd say 0.3s is probably the sweet spot

maxwowo:
> Also what do you think about hiding the course name altogether on hover? Would this be too confusing for users?

Do you mean hiding the course code and the class type? I did some hack to try and produce that effect but it didn't look very nice since the course code would constantly disappear and spawn back in the card which looks quite messy, it's also a bit confusing as well 

https://imgur.com/NuZM3dr

lukefisklennon:
> > Also what do you think about hiding the course name altogether on hover? Would this be too confusing for users?
> 
> Do you mean hiding the course code and the class type? I did some hack to try and produce that effect but it didn't look very nice since the course code would constantly disappear and spawn back in the card which looks quite messy, it's also a bit confusing as well since you might forget what course / class type you're looking at and will end up having to unhover just to check that
> 
> https://imgur.com/NuZM3dr

Although I think the transition could be made to work, the issue about it being confusing to users as you say can't really be solved. So keeping the course name always visible sounds good.

maxwowo:
This is looking really good 🔥 However I noticed that when you drop a course, the card immediately goes back to the unhovered mode, which is quite strange since the cursor should be right on top of it in order to drop it in the first place. Did you notice this when you were implementing this issue? Would be great if the card could stay hovered after being dropped

https://imgur.com/jB9zmHn

itsViggo:
> This is looking really good 🔥 However I noticed that when you drop a course, the card immediately goes back to the unhovered mode, which is quite strange since the cursor should be right on top of it in order to drop it in the first place. Did you notice this when you were implementing this issue? Would be great if the card could stay hovered after being dropped
> 
> https://imgur.com/jB9zmHn

hmmm yeah that's strange, i'll try and fix it.

PR #116, opened 2020-06-29
Feature/53 move inventory to side

lukefisklennon:
This PR moves the inventory to the right side of the timetable, and hides it when not in use. Known issue: inventory classes won't be in the order you dragged them in (not a new problem), but this will be fixed while improving DnD.


client/src/interfaces/DbCourse.tsx
```diff
@@ -81,31 +86,32 @@ const dbTimesToPeriod = (dbTimes: DbTimes): Period => ({
...
+      course: courseData,
       activity: dbClass.activity,
       periods: dbClass.times.map(dbTimesToPeriod),
       enrolments: dbClass.courseEnrolment.enrolments,
       capacity: dbClass.courseEnrolment.capacity,
     };
-    classData.periods.forEach((period) => {
-      latestClassFinishTime = latestClassFinishTime > period.time.end
-        ? latestClassFinishTime : period.time.end;
+    classData.periods!.forEach((period) => {
```
maxwowo, line 105: Do we need this `!.` between the `periods` and `forEach`? Would `.` be sufficient?

client/src/interfaces/DbCourse.tsx
```diff
@@ -81,31 +86,32 @@ const dbTimesToPeriod = (dbTimes: DbTimes): Period => ({
  * const courseInfo = dbCourseToCourseData(json)
  */
 export const dbCourseToCourseData = (dbCourse: DbCourse): CourseData => {
-  const classes: Record<string, ClassData[]> = {};
-  let latestClassFinishTime = 0;
+  const courseData = {
```
maxwowo, line 89: Let's give this variable the type `CourseData`. Although it works without the type, it's generally good to type object variables to make use of IDE auto-suggest and enforce the types of fields

client/src/interfaces/DbCourse.tsx
```diff
@@ -81,31 +86,32 @@ const dbTimesToPeriod = (dbTimes: DbTimes): Period => ({
...
+    code: dbCourse.courseCode,
+    name: dbCourse.name,
+    activities: {} as Activities,
```
maxwowo, line 92: No need for `as Activities` if we typed this object

client/src/components/timetable/TimetableLayout.tsx
```diff
@@ -9,10 +9,28 @@ const headerPadding = 15;
 const BaseCell = styled.div<{
   x: number
   y: number
+  endX?: boolean
+  endY?: boolean
```
maxwowo, line 13: What are these two props used for?

client/src/components/timetable/Timetable.tsx
```diff
@@ -42,12 +53,30 @@ const Timetable: FunctionComponent<TimetableProps> = ({
...
     ), defaultEndTime) - defaultStartTime}
   >
+    <SelectWrapper>
+      <CourseSelect
+        selectedCourses={selectedCourses}
+        assignedColors={assignedColors}
+        handleSelect={onSelectCourse}
+        handleRemove={onRemoveCourse}
+      />
+    </SelectWrapper>
```
maxwowo, line 72: Does it make sense to have the timetable nest the course select? I've always thought of the two components as different entities responsible for different things without too much overlap between them. I reckon timetable should just contain the components which relate directly to the timetabling experience, like dropzones, timetable layout, inventory etc. What do you think?

client/src/components/inventory/Inventory.tsx
```diff
@@ -1,51 +1,92 @@
...
+      canDrop: monitor.canDrop(),
+      isOver: monitor.isOver(),
+    }),
+  });
+
+  return (
+    <StyledInventory ref={drop}>
+      {classNodes}
+      <StyledInventoryText>
+        Drag classes here which you want out of the way
```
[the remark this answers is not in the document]
lukefisklennon: Sounds good

client/src/api/getCoursesList.ts
```diff
@@ -1,6 +1,15 @@
 import { CoursesList } from '../interfaces/CourseOverview';
 import { API_URL } from './config';
 
+const toCoursesList = (data: any): CoursesList => (
+  data.map((course: any) => ({
```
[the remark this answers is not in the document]
lukefisklennon: Yup ok

client/src/components/CourseSelect.tsx
```diff
@@ -50,6 +50,8 @@ let fuzzy = new Fuse<CourseOverview, SearchOptions>([], searchOptions);
 const StyledSelect = styled(Box)`
   width: 100%;
   text-align: left;
+  position: relative;
+  left: -1px;
```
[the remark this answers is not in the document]
lukefisklennon: I'm not 100% sure why, but it could be something to do with the timetable borders being box shadows. (Before just the inner borders were box shadows, but now the outer borders too because now that the inventory has been added, having a border around the whole grid component wouldn't work.) This is just a theory though, there might be something else causing it.

client/src/interfaces/DbCourse.tsx
lukefisklennon: Ahh yeah, that was due to something else I didn't end up going with, I'll take it out now

client/src/interfaces/DbCourse.tsx
lukefisklennon: For some reason I thought it was needed, but yeah it works without that

client/src/components/timetable/TimetableLayout.tsx
lukefisklennon: They're used to signal to `BaseCell` whether or not that cell is the last cell in that row/column. It's so that the cell knows what rounded corners to apply. It used to be that the rounded corners were on the entire grid component, but now that the inventory is included, that doesn't work.

client/src/components/timetable/Timetable.tsx
lukefisklennon: I had nested them before because it needed to be in the grid layout, due to the inventory protruding off to the side. However I don't think that's needed now, so I'll move it back out.

client/src/components/CourseSelect.tsx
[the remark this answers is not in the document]
maxwowo: I see, that's alright, let's just leave it there for now. Can you keep this at the back of your mind though when you're working on the grid and DND issue? The less bandaid styles we have the better

client/src/interfaces/DbCourse.tsx
maxwowo: Yep I think you'd need that there if the `courseData` object wasn't typed, but because it's typed now the TS language server should know what the type of `activities` should be

client/src/components/CourseSelect.tsx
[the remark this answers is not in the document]
lukefisklennon: Yea agreed, I'll sort it out in the next PR

PR #118, opened 2020-06-30
feature/31 create graphical error notifications

itsViggo:


lukefisklennon:
Looks good! There's just a couple of minor things, like do you reckon the contrast with the background is high enough here?
image
Also I noticed a weird glitch where when you close the notification, the text disappears just before the notification disappears (only for like one frame). 6x slowed down gif of it happening:
notangles3
These are only very minor things though, we probably don't need to fix them now.

client/src/api/getCourseInfo.ts
```diff
@@ -18,19 +19,21 @@ const getCourseInfo = async (
...
   term: string,
   courseCode: string,
-): Promise<CourseData | null> => {
+): Promise<CourseData | NetworkError> => {
   const baseURL = `${API_URL}/terms/${year}-${term}`;
   try {
     const data = await fetch(`${baseURL}/courses/${courseCode}/`);
+    if (data.status === 400) {
+      return { message: 'Internal server error' };
+    }
```
maxwowo, line 28: Instead of returning a error object, what do you think about throwing a custom error? In JS you can define customized errors like 
```javascript
class NetworkError extends Error {
  constructor(message = '') {
    super(message);
  }
}
```
and instead of `line 26`, we can write something like 
```javascript
if (data.status === 400) {
  throw new NetworkError('Internal server error');
}
```
 so that when you call `getCourseInfo()` you can do something like
```javascript
try {
  const selectedCourseClasses = await getCourseInfo(
    '2020',
    'T2',
    courseCode,
  );
  const newSelectedCourses = [...selectedCourses, selectedCourseClasses];
  populateTimetable(selectedCourseClasses); // TODO: temp until auto-timetabling is done
  setSelectedCourses(newSelectedCourses);
} catch (e) {
  if (e instanceof NetworkError) {
    setErrorMsg(selectedCourseClasses.message);
  }
}
```

client/src/api/getCoursesList.ts
```diff
@@ -15,14 +16,16 @@ import { API_URL } from './config';
...
+    const data = await timeoutPromise(1000, fetch(`${baseURL}/courses/`));
+    if (data.status === 400) {
+      return { message: 'Internal server error' };
+    }
     return data.json();
   } catch (error) {
-    console.error(error);
-    return null;
+    return { message: 'Could not connect to server' };
   }
```
maxwowo, line 29: We could use the error handling technique I mentioned before for this section of the code as well

client/src/api/config.ts
```diff
@@ -13,3 +13,21 @@ const API_CONFIG: Record<string, string> = Object.freeze({
...
+        clearTimeout(timeoutId);
+        resolve(res);
+      },
+      (err) => {
+        clearTimeout(timeoutId);
+        reject(err);
+      },
+    );
+  });
+}
```
maxwowo, line 33: This shouldn't be in `config.ts` since it's not a configuration, can you make a new file in `client/utils/` and put this code in there instead?

maxwowo:
> Looks good! There's just a couple of minor things, like do you reckon the contrast with the background is high enough here?
> image
> Also I noticed a weird glitch where when you close the notification, the text disappears just before the notification disappears (only for like one frame). 6x slowed down gif of it happening:
> notangles3
> These are only very minor things though, we probably don't need to fix them now.

Sharp eyes! I think the reason the text disappears first before the snackbar goes away is because 

1) There's a single snackbar 

2) Whether the snackbar is open or not depends on if the error message it contains is empty or not, `empty error message => close snackbar`

3) Closing the snackbar by clicking on the cross calls a handler **which sets the error message it contains to be empty**, which means the snackbar will close. So the text will always disappear first before the snackbar closes, since it's the precondition for the snackbar to close

I reckon we can fix this issue by using separate variables for snackbar text and error messages

lukefisklennon:
@maxwowo Ahh I see, that makes sense. I'm probably being a bit pedantic, unlikely that anyone will notice. But if it's an easy fix, might as well.

maxwowo:
For some reason when I hover over the close button the snackbar would shrink a little bit, any idea what might be causing this? 
gif

client/src/api/getCourseInfo.ts
```diff
@@ -18,19 +20,21 @@ const getCourseInfo = async (
...
-): Promise<CourseData | null> => {
+): Promise<CourseData> => {
   const baseURL = `${API_URL}/terms/${year}-${term}`;
   try {
-    const data = await fetch(`${baseURL}/courses/${courseCode}/`);
+    const data = await timeoutPromise(1000, fetch(`${baseURL}/courses/${courseCode}/`));
+    if (data.status === 400) {
+      throw new NetworkError('Internal server error');
+    }
+    console.log();
```
maxwowo, line 30: Don't need this

client/src/api/getCoursesList.ts
```diff
@@ -29,14 +31,16 @@ const toCoursesList = (data: FetchedCourse[]): CoursesList => (
...
   term: string,
-): Promise<CoursesList | null> => {
+): Promise<CoursesList> => {
   const baseURL = `${API_URL}/terms/${year}-${term}`;
   try {
-    const res = await fetch(`${baseURL}/courses/`);
-    return toCoursesList(await res.json());
+    const data = await timeoutPromise(1000, fetch(`${baseURL}/courses/`));
+    if (data.status === 400) {
+      throw new NetworkError('Internal server error');
```
maxwowo, line 39: Will this error ever be shown? Because whenever this line is executed, the error is caught by `line 42`, that then throws another error which shadows the error on this line

We could solve this by making the `timeoutPromise` function throw a custom "TimeoutError" or something and checking for that as well as `NetworkError` in the catch block in `line 42`

PR #122, opened 2020-07-14
Feature/131 clash notification

esyw:
Basic notification of clashing classes by adding a small red border to class cards


client/src/App.tsx
```diff
@@ -120,6 +122,36 @@ const App: FunctionComponent = () => {
...
+              }
+            }
+          });
+        });
+      });
+    });
+    console.log('all classes with clashes:', newClashes);
+    return newClashes;
+  }
+  // checkClashes();
```
maxwowo, line 153: No need for this comment

client/src/App.tsx
```diff
@@ -120,6 +122,36 @@ const App: FunctionComponent = () => {
...
+              }
+              if (!newClashes.includes(classActivity2.id)) {
+                newClashes.push(classActivity2.id);
+              }
+            }
+          });
+        });
+      });
+    });
+    console.log('all classes with clashes:', newClashes);
```
maxwowo, line 150: No need for this log

client/src/App.tsx
```diff
@@ -120,6 +122,36 @@ const App: FunctionComponent = () => {
     ));
   };
 
+  function hasTimeOverlap(period1: ClassTime, period2: ClassTime): boolean {
```
maxwowo, line 125: Let's use arrow functions instead to be consistent with the rest of the codebase

client/src/App.tsx
```diff
@@ -120,6 +122,36 @@ const App: FunctionComponent = () => {
...
+    return ((period1.day === period2.day && period1.start >= period2.start
+        && period1.start < period2.end)
+     || (period1.day === period2.day && period2.start >= period1.start
+        && period2.start < period1.end));
+  }
+
+  function checkClashes():Array<String> {
```
maxwowo, line 132: Use arrow function

maxwowo, line 132: What if the clashes are represented as `Array<ClassData>` instead of `Array<String>`? It'll be more descriptive that way and since objects are pushed as references anyway there won't be any performance drawbacks. Then instead of comparing `classActivity.id` and so on, we can just compare the objects themselves.

esyw: decided to store as ClassPeriod instead such that the border only appears on the actual times there's a clash (vs all the times for that activity) - thought that might actually be clearer since you cant see what's underneath the overlapping blocks

PR #129, opened 2020-07-21
Feature/166 Add feedback forms

maxwowo:
This PR
- Adds a feedback button in the navbar which links to a Google form
- Changes "feedback" to "email" in the page footer
- Links Google form to about modal

image
image

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -164,7 +164,7 @@ const App: FunctionComponent = () => {
                 <br />
                 Made by &gt;_ CSESoc UNSW&nbsp;&nbsp;•&nbsp;&nbsp;
                 <Link target="_blank" href="mailto:projects@csesoc.org.au">
-                  Feedback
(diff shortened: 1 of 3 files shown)


lukefisklennon:
I'd suggest perhaps putting the "feedback" to the right of the dark theme button, so all the text buttons are grouped together. What do you reckon?

maxwowo:
> I'd suggest perhaps putting "feedback" to the right of the dark theme button, so all the text buttons are grouped together. What do you reckon?

Good point

maxwowo:
> perhaps there could also be a link to the form over the "please contact us" in the About?

Fair enough

esyw:
just noticed that on light mode, the feedback button text is black instead of white for me?

maxwowo:
> just noticed that on light mode, the feedback button text is black instead of white for me?

Oops, good spot 💪

PR #146, opened 2020-12-19
Feature/208 temporary term display

lukefisklennon:


client/src/components/Navbar.tsx
```diff
@@ -49,6 +49,11 @@ const NavButton = styled(Button)`
   margin-right: 20px;
 `;
 
+const Weak = styled.span`
+  font-weight: 300;
+  opacity: 0.8;
```
esyw, line 54: What do you think of adding left padding to match the gap between the logo and "Notangles", and/or decreasing the font to be a bit smaller than "Notangles"?

client/src/App.tsx
```diff
@@ -276,11 +286,15 @@ const App: FunctionComponent = () => {
             isDarkMode={isDarkMode}
             handleDrawerOpen={handleDrawerOpen}
           />
-          <FriendsDrawer
-            isFriendsListOpen={isFriendsListOpen}
-            isLoggedIn={isLoggedIn}
-            setIsLoggedIn={handleSetIsLoggedIn}
-          />
+          {
+            IS_PREVIEW && (
```
esyw, line 290: I might've missed it but should menu button in the navbar be hidden as well?

lukefisklennon: Ah good point

client/src/components/Navbar.tsx
lukefisklennon: Yep ok, what do you think now?

esyw: Might not even need the brackets with the difference in font to make it clear, but yeah looks good!

esyw (approved):

client/src/components/Navbar.tsx
maxwowo: Yeah I think it's just a bit cleaner without the brackets, not a big deal though

image
image

maxwowo (approved):

client/src/components/Navbar.tsx
lukefisklennon: Agreed, done now

PR #185, opened 2021-06-16
Fixed course start and end times

emilytangg:


oliverxu42 (approved):
Looks good!

lukefisklennon (changes requested):
Nice work! There's just one issue with how the start and end times are being rounded:

image

Because the end time is floored instead of ceiling'd, there's half a row less than what's needed. Same for the start time, so the floor and ceiling can just be swapped.

emilytangg:
> Nice work! There's just one issue with how the start and end times are being rounded:
> 
> image
> 
> Because the end time is floored instead of ceiling'd, there's half a row less than what's needed. Same for the start time, so the floor and ceiling can just be swapped.

Oops my bad, I will fix this :)

lukefisklennon (approved):

PR #192, opened 2021-09-08
[Oliver & Luke] Feature/class f2f online icons

lukefisklennon:
This PR adds icons to the dropzones which represent whether those class periods have in-person and/or online options available. It also fixes the issue where multiple dropzones overlap, which causes inconsistent shading.


oliverxu42 (approved):

martanman (approved):
good stuff

PR #202, opened 2022-02-13
Feature/add ics saving

Peedee2002:
Added ICS saving to notangles. I dont know how the design is supposed to look, so i literally just made the laziest frontend code i could. here is how it looks:
image
(notice the button that clearly says "generate ICS file".)
You can tell me how you would want it to look and ill be happy to code that up. Also, i had to add a new constant that needs to be maintained:
"firstMomentOfTerm".
This is a constant that tells me what is the UTC time of the midnight of the first monday of term. I need this to accurately calculate the time for calandars.
here is a sample of what happens if you have this example calander:
image

this pops up when you hit the button:
image
If you decide to open it, it immediately asks to add it to your calander:
image
This also shows what the events look like after getting generated.

Cheers!


Peedee2002:
updated this to deal with dumb merge conflicts

PR #204, opened 2022-02-26
added animated logo

gracexkan:


emilytangg:
I noticed that the size of the new logo is a bit smaller than the original. I'm just wondering if this was intentional or not :)  Otherwise the code looks good to me! 

Original:
image

New:
image

gracexkan:
> I noticed that the size of the new logo is a bit smaller than the original. I'm just wondering if this was intentional or not :) If yes, then it's all good. Looks good to me!
> 
> Original: image
> 
> New: image

o ye the problem was that the image that I edited to make the animation turned out to be a slightly different one (sizing issues :(() to the one that is actually being used on notangles but ye I'll update the css since it's very noticeable

emilytangg (approved):
Thanks for fixing the logo size, looks good!

munjoonteo (approved):

PR #210, opened 2022-03-05
Feature/ntgl 256 class selection

martanman:
Adds the arrow buttons to the sides of courses that let you choose through courses at the same timeslot.
Also changes drag just to not update classes too prematurely.


munjoonteo (approved):
berry nice

oliverxu42 (approved):
Good stuff!

PR #250, opened 2022-03-19
Fixed courses duplicating in course select

martanman:
Instead of filtering courses list which keeps duplicates it maps each unique class code to a course to avoid courses which have duplicates.


oliverxu42 (approved):
Lol you stole my ticket hahaha looks good though

munjoonteo (approved):

PR #283, opened 2022-03-27
Feature/ntgl 280 differentiate ug pg

oliverxu42:
Adds an indicator to show if the course is under/post grad.

image


client/src/components/CourseSelect.tsx
```diff
@@ -352,6 +357,7 @@ const CourseSelect: React.FC<CourseSelectProps> = ({ assignedColors, handleSelec
             </StyledIcon>
             <span>{option.code}</span>
             <Weak>{option.name}</Weak>
+            <Career>{option.career}</Career>
```
munjoonteo: What do you think about making it just UG or PG instead of "Undergraduate" and "Postgraduate"

munjoonteo (changes requested):

client/src/components/CourseSelect.tsx
oliverxu42: Yeah true I think UNSW abbreviates to UGRD/PGRD so that might be better.

munjoonteo (approved):

PR #288, opened 2022-03-28
Bugfix/ntgl-285-left-right-scroll-drag-broken

anjerraa:
- left and right scroll/drag was broken on mobile due to the div wrapping that I had around the timetable and the ics button (purpose was so the the button sticks to the bottom of the timetable in higher resolution screens)
- removed the div wrappings, ended up changing the min height of the content wrapper instead so the button doesnt have huge paddings in higher res
- Changed CourseSelect.tsx so that when in mobile view, the course name doesn't show up (avoid squished text)


munjoonteo:
@oliverxu42 can you test this on your big screen

munjoonteo (approved):

oliverxu42:
@munjoonteo fine on the ultra-wide. @anjerraa before the course-select would also scroll across to avoid it being squished...
image

anjerraa:
@oliverxu42 How do you replicate the screenshot you sent?

munjoonteo (approved):

PR #289, opened 2022-03-28
History (Undo, Redo, Reset)

martanman:


munjoonteo (changes requested):
1. Clicking the reset button adds an action every time i.e. click reset 5 times, have to undo 5 times. Would it be possible to make it so that no matter how many times you click reset you only have to undo once?
2.  When you drag a class around, every time the class enters a valid dropzone, an action is added to the list. So when you undo you kinda retrace the direction in which you dragged the class. Can an action be added to the list only after the class is dropped into a dropzone?

munjoonteo (approved):

PR #292, opened 2022-03-30
History

martanman:
with fixes for theme and breakpoints


oliverxu42 (approved):

client/src/components/CourseSelect.tsx
```diff
@@ -426,19 +427,21 @@ const CourseSelect: React.FC<CourseSelectProps> = ({ assignedColors, handleSelec
...
-              deleteIcon={<CloseRounded />}
-              {...getTagProps({ index })}
-              onDelete={() => {
-                setSelectedValue(selectedValue.filter((course) => course.code !== option.code));
-                handleRemove(option.code);
-              }}
-            />
-          ))
+          value.map((option: CourseOverview, index: number) => {
+            return (
```
munjoonteo: Why the explicit `return`?

munjoonteo (changes requested):

client/src/components/CourseSelect.tsx
martanman: no reason

PR #296, opened 2022-04-03
fix: Add api call to get timezone

Peedee2002:
this was probably the way i should have done this in the beginning. Now, the only thing the constants have to worry about is getting the first day of term. I couldnt figure out where i can fetch this data using the API so ill leave that for now. This should make dealing with my code much more manageable.


Peedee2002:
will reopen when i fix the spammed api thing

Peedee2002:
ok uncursified

munjoonteo (approved):

PR #306, opened 2022-04-08
Autotimetabling

munjoonteo:


auto_server/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.10.4-slim-buster
+
+WORKDIR /auto_server
+
+COPY requirements.txt requirements.txt
+
+RUN pip3 install -r requirements.txt
+
+COPY . .
+
```
lhjt, line 10: A port should be exposed by this image.

client/src/interfaces/Course.ts
```diff
@@ -42,3 +42,12 @@ export interface ClassTime {
...
 }
+
+export interface AutoData {
+  start: number;
+  end: number;
+  days: string;
+  gap: number;
+  maxdays: number;
+  periodsListSerialized: string;
+}
```
lhjt: Consider adding the missing newline at the end of this file (run prettier on all files)

server/src/controllers/getAutotimetable.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as grpc from '@grpc/grpc-js';
+import { TimetableConstraints } from '../proto/autotimetabler_pb';
+import { AutoTimetablerClient } from '../proto/autotimetabler_grpc_pb';
+
+
+type AutoData = TimetableConstraints.AsObject
+
+
+export const getAuto = async (req: Request, res: Response) => {
+  var client = new AutoTimetablerClient('localhost:50051', grpc.credentials.createInsecure());
```
lhjt: The address that the server connects to should be specified with an environment variable. Setting it to localhost will not work in the production cluster.

server/src/database.ts
```diff
@@ -1,162 +1,167 @@
-import { MongoClient, Db } from 'mongodb';
```
lhjt, line 1: Why is this file being kept if it's just being fully commented out?

lhjt (changes requested):

lhjt (commented):
This looks good to me so far. I'll approve this so that we can merge once I have configured the CD component for the automatic timetabler server.

server/src/controllers/getAutotimetable.ts
lhjt: In the future, you may want to consider just having the entire connection string as an env var, but this may do for now,

server/src/config.ts
```diff
@@ -1,29 +1,18 @@
...
-
-  [Env.STAGING]: {
-    database: 'mongodb://database.notangles-db:27017/',
-    scraper: 'mongodb://localhost:27017',
-  },
-
   [Env.PROD]: {
-    database: 'mongodb://database.notangles-db:27017/',
-    scraper: 'mongodb://database.notangles-db:27017',
+    auto: 'localhost:50051',
```
lhjt: This should be using an environmental variable instead (such as `AUTO_SERVER_HOSTNAME`)

lhjt (changes requested):

lhjt (approved):

PR #307, opened 2022-04-08
Ntgl 222 sentry setup

OAK-02:
set-up sentry for frontend error monitoring and maodified smaple tracing rate for production


oliverxu42:
ooop please fix conflicts as soon as you can :)

PR #364, opened 2022-04-21
gave algo objective to maximize enforced constraints 

martanman:
Affects days of the week of classes (i.e mad days of uni) and earliest start times and latest end time constraints


munjoonteo:
I should've pointed this out before but for multiplying everything by 2, can you make that a constant instead of a magic number? And explain in a comment that you need to scale it to get rid of half hours.

Also why do you need to multiply by 100 on lines 123 and 124?

martanman:
@munjoonteo i * 100 indicates the start of day i, as time are given as DotW * 100 + timeOfDay.
ok i will make 2 a constant (even though nothing is a constant in python but yes if u mean big capital letters global variable then oki)

munjoonteo:
Oh that makes sense ok

munjoonteo:
Do you think we should let the user know that not all of their constraints have been satisfied? Maybe a yellow coloured alert or something

martanman:
@munjoonteo I've considered it, and possibly, but not totally sure, and not sure what exactly the message should say. it _may_ in fact be implicit (or may not) that the autotimetabler does the best it can, or perhaps even more likely, it may not even be conceptualised as constraint-based by the user.
would the average person expect it to strictly adhere to their constraints?

but yeah sure tbh.

munjoonteo:
Hmm I guess it depends on how blatant the violation is right... like say if I said no earlier than 12pm and it gives me an 11am class then sure I could live with that. But if I say no Fridays and it gives me a Friday class I'm like ????

I guess the message could say something along the lines of "Not all requirements satisfied"?

client/src/components/Alerts.tsx
```diff
@@ -19,7 +19,7 @@ const Alerts: React.FC = () => {
   };
 
   const getAutoSeverity = () => {
-    return alertMsg === 'Success!' ? 'success' : 'error';
+    return alertMsg === 'Success!' ? 'success' : alertMsg.startsWith('Could not') ? 'warning' : 'error';
```
munjoonteo: At this point it's probably more readable to have it as a switch statement right?

munjoonteo (changes requested):

client/src/components/Alerts.tsx
martanman: i _was_ considering it but lowkey switch statements are yucky boilerplate. plus it would mean having to write out the full message instead of having the string.startsWith()

munjoonteo (approved):

PR #439, opened 2022-06-21
NTGL-301: display only classes which are open for enrolment

anjerraa:
- Changed the `isHideFullClasses` toggle to `isShowOnlyOpenClasses` because there aren't just open and full classes, but also on-hold, closed and tentative(?).
- Previously, on-hold classes would show up on the timetable with `0/capacity`, but now it will have the same warning icon as full classes, but instead of enrolment number it would show **On Hold**. This is to make it clearer that the class is on-hold rather than 0 enrolments.
- Added `Status` as a an union type and as a field to `DbClass` and `ClassData` types.


anjerraa:
Removed "Closed" and "Tentative" from `Status` because apparently there are no classes with such statuses.

munjoonteo:
image

JURD7251 doesn't show the Monday 12pm lecture because it's on hold but I didn't turn on the setting. http://timetable.unsw.edu.au/2022/JURD7251.html

anjerraa:
> image
> 
> JURD7251 doesn't show the Monday 12pm lecture because it's on hold but I didn't turn on the setting. http://timetable.unsw.edu.au/2022/JURD7251.html

When you check the link, there is another "Open" lecture stream which is stream B and that one doesn't even show up on the timetable though, just stream A?

oliverxu42:
> image
> 
> JURD7251 doesn't show the Monday 12pm lecture because it's on hold but I didn't turn on the setting. http://timetable.unsw.edu.au/2022/JURD7251.html

hmmmm I think there are bigger issues with JURD7251 - like the Monday 12 pm shows up when you drop the lecture from unscheduled...  

image

PR #465, opened 2022-07-15
Feature/ntgl 334 consolidate overlapping period info into one period

gracexkan:
Before:

Screen Shot 2022-07-15 at 11 33 04 pm
Screen Shot 2022-07-15 at 11 33 08 pm

After:

Screen Shot 2022-07-15 at 11 36 56 pm


munjoonteo:
image
What is this o.o

client/src/api/getCourseInfo.ts
```diff
@@ -16,6 +17,51 @@ import timeoutPromise from '../utils/timeoutPromise';
  * @example
  * const selectedCourseClasses = await getCourseInfo('2019', 'T1', 'COMP1511')
  */
+
+const convertTimesToList = (dbClassWeeks: string, dbClassTimesList: number[]): number[] => {
+  for (let k = 0; k < dbClassWeeks.length; k++) {
```
munjoonteo, line 21: This reads a lot like C - in Python I would write this entire loop with the `enumerate` keyword. Can you find an equivalent in JS?

client/src/api/getCourseInfo.ts
```diff
@@ -16,6 +17,51 @@ import timeoutPromise from '../utils/timeoutPromise';
...
+        dbClassTimesList.push(parseInt(dbClassWeeks[k]));
+      }
+    }
+  }
+  return dbClassTimesList;
+};
+
+const sortUnique = (arr: number[]): number[] => {
+  if (arr.length === 0) return arr;
+  arr = arr.sort(function (a, b) {
```
munjoonteo: `const` syntax for functions please

client/src/api/getCourseInfo.ts
```diff
@@ -24,6 +70,59 @@ const getCourseInfo = async (year: string, term: string, courseCode: CourseCode)
       throw new NetworkError('Internal server error');
     }
     const json: DbCourse = await data.json();
+    json.classes.forEach((dbClass) => {
+      if ('times' in dbClass && dbClass.times.length > 1) {
+        for (let i = 0; i < dbClass.times.length - 1; i += 1) {
```
munjoonteo, line 85: Given the context of this PR I understand what this code is doing but someone reading this for the first time would probably have no idea why this is necessary. Can you add a comment?

client/src/api/getCourseInfo.ts
```diff
@@ -24,6 +70,59 @@ const getCourseInfo = async (year: string, term: string, courseCode: CourseCode)
...
+          for (let j = i + 1; j < dbClass.times.length; j += 1) {
+            let dbClassTimesOne = dbClass.times[i];
+            let dbClassTimesTwo = dbClass.times[j];
+            if (
+              dbClassTimesOne.day == dbClassTimesTwo.day &&
```
munjoonteo: This could probably be a function

client/src/api/getCourseInfo.ts
```diff
@@ -16,6 +17,51 @@ import timeoutPromise from '../utils/timeoutPromise';
...
+        // this week is a two digit number and the two digits must be added together
+        dbClassTimesList.push(parseInt(dbClassWeeks[k] + dbClassWeeks[k + 1]));
+        k++;
+      } else {
+        // this week is not a two digit number so we just add the one digit to it
+        dbClassTimesList.push(parseInt(dbClassWeeks[k]));
+      }
+    }
+  }
+  return dbClassTimesList;
```
munjoonteo: Javascript objects are always (ok well not technically) passed by reference. You don't need to return the new list.

client/src/api/getCourseInfo.ts
```diff
@@ -16,6 +17,51 @@ import timeoutPromise from '../utils/timeoutPromise';
...
+};
+
+const sortUnique = (arr: number[]): number[] => {
```
munjoonteo, line 56: You should've copied this so thread instead https://stackoverflow.com/questions/1960473/get-all-unique-values-in-a-javascript-array-remove-duplicates XD

client/src/api/getCourseInfo.ts
```diff
@@ -24,6 +70,59 @@ const getCourseInfo = async (year: string, term: string, courseCode: CourseCode)
...
+
+              dbClassTimesList = convertTimesToList(dbClassTimesOne.weeks, dbClassTimesList);
+              dbClassTimesList = convertTimesToList(dbClassTimesTwo.weeks, dbClassTimesList);
+
+              dbClassTimesList = sortUnique(dbClassTimesList);
+
+              let newWeeks: string = '';
+              let rangeStart = false;
+
+              for (let k = 0; k < dbClassTimesList.length; k++) {
```
munjoonteo, line 100: This can also use the same `enumerate` replacement you created in `convertTimesToList`

client/src/api/getCourseInfo.ts
```diff
@@ -24,6 +70,59 @@ const getCourseInfo = async (year: string, term: string, courseCode: CourseCode)
...
+                  rangeStart = true;
+                }
+              }
+              dbClassTimesOne.weeks = newWeeks;
+              dbClass.times.splice(dbClass.times.indexOf(dbClassTimesTwo), 1);
+            }
+          }
+        }
+      }
+    });
```
munjoonteo, line 130: New line here would be nice

munjoonteo (changes requested):

client/src/api/getCourseInfo.ts
gracexkan: @munjoonteo I've used a for loop here because I do k++ inside the loop sometimes, can the equivalent still do that?

oliverxu42 (approved):

munjoonteo (approved):

PR #515, opened 2022-08-19
Improve drag behavior regarding clashes

martanman:
fixes the width change of a card when it is placed on a clashing zone not having a transition but instead being instantaneous. It now transitions between the two widths as the already dropped classes do.

gets rid of behavior wherein the dragged card will stutter between two dropzones

Peek 2022-08-19 21-54

by making the dropzone calculation invariable from current card width size.

Now, width transition of the dragging card will shrink/grow towards/from the middle of the card rather than to/from the left side of the card.

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -169,7 +169,15 @@ const App: React.FC = () => {
       )
     ).then((result) => {
       const addedCourses = result.filter((course) => course.code !== undefined) as CourseData[];
-      const newSelectedCourses = [...selectedCourses, ...addedCourses];
+
+      // prevent duplication on recompile
+      const newSelectedCoursesWithDups = [...selectedCourses, ...addedCourses];
+      const newSelectedCourseCodes = newSelectedCoursesWithDups.map((course) => course.name);
+      const newSelectedCourses = newSelectedCoursesWithDups.filter(
+        (course, index) =>
+          newSelectedCourseCodes.indexOf(course.name) === newSelectedCourseCodes.lastIndexOf(course.name) ||
+          newSelectedCourseCodes.indexOf(course.name) === index
+      );
 
       setSelectedCourses(newSelectedCourses);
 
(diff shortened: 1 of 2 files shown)


client/src/App.tsx
```diff
@@ -169,7 +169,15 @@ const App: React.FC = () => {
...
     ).then((result) => {
       const addedCourses = result.filter((course) => course.code !== undefined) as CourseData[];
-      const newSelectedCourses = [...selectedCourses, ...addedCourses];
+
+      // prevent duplication on recompile
+      const newSelectedCoursesWithDups = [...selectedCourses, ...addedCourses];
+      const newSelectedCourseCodes = newSelectedCoursesWithDups.map((course) => course.name);
+      const newSelectedCourses = newSelectedCoursesWithDups.filter(
+        (course, index) =>
+          newSelectedCourseCodes.indexOf(course.name) === newSelectedCourseCodes.lastIndexOf(course.name) ||
```
munjoonteo, line 178: What does this do

client/src/utils/Drag.ts
```diff
@@ -408,6 +422,18 @@ window.addEventListener('scroll', onScroll, { passive: false });
 
 let eventId = '';
 
+let prevWidth: number = -1;
+
+let resizeObserver: ResizeObserver = new ResizeObserver((entries: ResizeObserverEntry[]) => {
+  const entryWidth = entries[0].contentRect.width;
```
munjoonteo, line 428: According to this https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver this might not work in Firefox ? Or is it referring to something else ?

munjoonteo (changes requested):

client/src/App.tsx
martanman: in python pseudocode

`courses = list(set(courses))`

client/src/utils/Drag.ts
martanman: it says

> Firefox implements `contentBoxSize` as a single content rect, rather than an array

but I'm accessing `contentRect` not `contentBoxSize` so that's not it. plus if it were `Array<T> | T` would ts really just let me compile 🤷‍♂️

munjoonteo (approved):

PR #516, opened 2022-08-21
Fix for navbar 0000 bug

Rayahhhmed:
When the user first gets on the site: 
1. they will by default have the 0000 data 
2. then from there onwards this data will be stored in a localstorage. 

The next time the user logs in, their data from localstorage would be used rather than the default 0000 data and then a fetch would be called to the timetable scraper to ensure that the data that is received is always updated throughout.


client/src/constants/timetable.ts
```diff
@@ -6,11 +6,22 @@ import timeoutPromise from '../utils/timeoutPromise';
...
-  let term = `T${termNumber}`;
+  let termData = {
+    year: '',
+    term: '',
+    termNumber: '',
+    termName: '',
+    firstDayOfTerm: '',
+  };
+
+  if (localStorage.getItem('termData')) {
```
munjoonteo: There is a localStorage API defined in `storage.ts`. Use that instead. Remember to add the required stuffs in `defaults.ts` as well.

client/src/context/AppContext.tsx
```diff
@@ -115,6 +115,16 @@ export const AppContext = createContext<IAppContext>({
 });
 
 const AppContextProvider = ({ children }: AppContextProviderProps) => {
+  let termData = {
```
munjoonteo, line 137: Is it necessary to initialise it like this? When it's already being done in App?

munjoonteo (changes requested):

PR #524, opened 2022-08-26
Ntgl 303 merge fix

Rayahhhmed:
Created a new branch from dev and fixed yotl problem.


munjoonteo:
raiyan i am sorry but we realised after u left today that ur still gonna have to call the API and only use localstorage as a fallback because if the term changes we can't just grab from localstorage.

i.e. 
1) make api call
2) if success: use data from api call, save data in localstorage
3) if failure: use data from localstorage

PR #526, opened 2022-08-27
Bugfix/ntgl 356 allow less than one hour event

manhualu:
- Custom event can now be shorter than one hour
- Made event card only show location if it is longer than one hour 
- Changed TimePicker to allow selection of minutes
- Fixed another bug where timetable breaks when user selects time with minutes that is before earliestStartTime or after latestEndTime


client/src/components/controls/CustomEvent.tsx
```diff
@@ -110,8 +110,8 @@ const CustomEvent: React.FC = () => {
       [uuid]: newEvent,
     });
 
-    setEarliestStartTime(Math.min(earliestStartTime, startTime.getHours()));
-    setLatestEndTime(Math.max(latestEndTime, endTime.getHours()));
+    setEarliestStartTime(Math.min(Math.ceil(earliestStartTime), Math.ceil(startTime.getHours() + startTime.getMinutes() / 60)));
```
munjoonteo: Should this be `Math.floor()` instead? Since you have to round down to get the earliest time.

client/src/components/controls/CustomEvent.tsx
```diff
@@ -212,10 +211,10 @@ const CustomEvent: React.FC = () => {
           <StyledListItem>
             <StyledListItemText primary="End time" />
             <TimePicker
-              views={['hours']}
               value={endTime}
               renderInput={(params) => {
-                const tooEarly = startTime.getHours() >= endTime.getHours();
+                const tooEarly =
```
munjoonteo: I noticed that this function is used in ExpandedEventView as well. You could put it into another folder and export it.

client/src/utils/Drag.ts
```diff
@@ -178,15 +178,12 @@ const timetableBorderWidthPlusOne = 2;
...
 const getIntersectionArea = (drag: DOMRect, drop: DOMRect) => {
-  // We pretend the drag element is its full width (i.e. the width of the drop rectangle)
-  // to avoid consecutive best-dropzone calculations returning different dropzones
-  // (wherein it will switch to-and-fro between two dropzones causing the card to stutter)
-  const newLeft = drag.left - (drop.width - drag.width) / 2;
-  const left = Math.max(newLeft, drop.left);
-  const right = Math.min(newLeft + drop.width, drop.right);
-  const bottom = Math.min(drag.bottom + timetableBorderWidthPlusOne, drop.bottom);
-  const top = Math.max(drag.top - timetableBorderWidthPlusOne, drop.top);
+  const left = Math.max(drag.left, drop.left);
```
munjoonteo: Any reason for reverting this code?

munjoonteo (changes requested):

client/src/components/timetable/ExpandedEventView.tsx
```diff
@@ -27,8 +27,8 @@ const ExpandedEventView: React.FC<ExpandedEventViewProps> = ({ eventPeriod, popu
 
   const [newName, setNewName] = useState<string>(name);
   const [newDays, setNewDays] = useState<Array<string>>([daysShort[day - 1]]);
-  const [newStartTime, setNewStartTime] = useState<Date>(new Date(2022, 0, 0, start));
-  const [newEndTime, setNewEndTime] = useState<Date>(new Date(2022, 0, 0, end));
+  const [newStartTime, setNewStartTime] = useState<Date>(new Date(2022, 0, 0, start, (start - Math.floor(start)) * 60));
```
munjoonteo: This date constructor could be a separate function as it's used four times in the same way.

munjoonteo (changes requested):

munjoonteo (approved):

PR #539, opened 2022-09-02
fixed bug where expanded view can't be accessed via mobile

gracexkan:
- Added onTouchStart since onMouseOver doesn't change anything on mobile


client/src/components/timetable/DroppedClass.tsx
```diff
@@ -161,7 +161,10 @@ const DroppedClass: React.FC<DroppedClassProps> = ({
       <StyledCard
         ref={element}
         onMouseDown={onDown}
-        onTouchStart={onDown}
+        onTouchStart={() => {
+          onDown
```
munjoonteo: I think you need to actually call the onDown function and pass in the event

munjoonteo (changes requested):

munjoonteo (approved):

PR #540, opened 2022-09-02
added scrolling functionality

gracexkan:
- added scroll for dropped classes so all classes can be seen in the dropzone


munjoonteo:
This is way too cursed can you make the actual unscheduled column scrollable instead?

gracexkan:
> This is way too cursed can you make the actual unscheduled column scrollable instead?

I had a look at it but it's the only way right now to make it scrollable since that column is a part of the big table that includes the actual timetable. Is there a way to only make a specific column in a table scrollable?

munjoonteo:
> > This is way too cursed can you make the actual unscheduled column scrollable instead?
> 
> I had a look at it but it's the only way right now to make it scrollable since that column is a part of the big table that includes the actual timetable. Is there a way to only make a specific column in a table scrollable?

Let's do some research 👀

PR #542, opened 2022-09-03
Feature/ntgl 353 clone custom event

sssijin:
this branch allows users to clone an event that is already created

- is line 121 - 154 in ExpandedEventView too similar to the multiple event stuff in CustomEvents? do I need to store it as a function in somewhere else and then export it...


munjoonteo:
Can you make it so that if there's an event on multiple days, all the days are highlighted in the day picker? i.e. link the events that are created together.

Also idk if this is skill diff or not but I accidentally clicked the delete event button instead of save and I was sad.

munjoonteo:
Actually I take back the first point I said. But I find it a bit odd that after you clone an event and you click on it again, the days doesn't change.

oliverxu42:
> 

Thoughts on making the "save changes" button at the bottom? Would be more consistent with the create event ui and avoid accidentally clicking delete. 
image

munjoonteo:
> Thoughts on making the "save changes" button at the bottom? Would be more consistent with the create event ui and avoid accidentally clicking delete.

We could actually just remove the delete button from the editing screen and only show that button when you first click on an event. i.e. the only thing you can do in the edit menu is save or cancel. 

Moving the save button to the bottom would also be good.

client/src/components/timetable/ExpandedEventView.tsx
```diff
@@ -54,6 +55,14 @@ const ExpandedEventView: React.FC<ExpandedEventViewProps> = ({ eventPeriod, popu
     setIsChanged(true);
   };
 
+  const timePickerStart = (isChanged: boolean, newStartTime: Date, start: number) => {
```
munjoonteo, line 66: What do these functions do? Can you add a JSDoc comment

client/src/components/timetable/ExpandedEventView.tsx
```diff
@@ -103,14 +118,52 @@ const ExpandedEventView: React.FC<ExpandedEventViewProps> = ({ eventPeriod, popu
...
+      updatedEventData[newEvent.event.id] = newEvent;
+    }
+
+    setCreatedEvents(updatedEventData);
+    setIsEditing(false);
+  }
+
+  const createEvent = (day: string) => {
+    const uuid = uuidv4();
+    const newEvent: EventPeriod = {
```
munjoonteo: This function looks identical to the start of `createEvent` in `CustomEvent.tsx`. You can factor out the code to create a new `EventPeriod` to a function.

munjoonteo (changes requested):

client/src/components/timetable/ExpandedEventView.tsx
sssijin: donee

client/src/components/timetable/ExpandedEventView.tsx
sssijin: donee

sssijin:
> We could actually just remove the delete button from the editing screen and only show that button when you first click on an event. i.e. the only thing you can do in the edit menu is save or cancel.
> 
> Moving the save button to the bottom would also be good.

Have done both of the above

munjoonteo (approved):

PR #557, opened 2022-09-13
fix cards jumping off of screen on drag start (on mobile / when screen scrolling)

martanman:


munjoonteo (approved):

munjoonteo:
smh it's legacy code ok

PR #574, opened 2022-09-23
reset popover values

sssijin:
resets create event values when create event popover is closed


client/src/components/controls/CustomEvent.tsx
```diff
@@ -87,17 +97,11 @@ const CustomEvent: React.FC = () => {
       const newEvent = createEvent(day);
       newEvents[newEvent.event.id] = newEvent;
     }
-
+
```
munjoonteo: ayo extra white space here ?

munjoonteo (changes requested):

munjoonteo (approved):

PR #575, opened 2022-09-24
Feature/friend support

Rayahhhmed:
The endpoints for the backend are done and can be checked with API documentation on confluence.


server/src/app.module.ts
```diff
@@ -1,23 +1,21 @@
...
 import { DatabaseModule } from './database/database.module';
+import { UserModule } from './user/user.module';
+import { FriendModule } from './friend/friend.module';
 
 @Module({
   imports: [
     ConfigModule.forRoot(), // so that we can pull in config
     AuthModule,
-    SumModule,
-    // AutoModule,
```
munjoonteo, line 16: o.o

server/src/auth/auth.controller.ts
```diff
@@ -8,34 +9,35 @@ import { LoginGuard } from './login.guard';
 export class AuthController {
   constructor(private readonly authService: AuthService) {}
 
-  @UseGuards(LoginGuard)
   @Get('/login')
+  @UseGuards(LoginGuard)
```
munjoonteo: Nitpick here: all the examples I've seen have the GET/POST etc. decorator as the last line

server/src/auth/auth.service.ts
```diff
@@ -3,30 +3,88 @@ import { InjectModel } from '@nestjs/mongoose';
...
+        },
+      );
     }
   }
 
-  async getUser(uidGiven: string): Promise<UserInterface | null> {
-    const response = await this.userModel.find({ google_uid: uidGiven });
-    return response.length !== 0 ? response.at(0) : null;
+  async getUser(uidGiven: string): Promise<UserInterface> {
+    const response = await this.userModel.findOne({ google_uid: uidGiven });
```
munjoonteo: I noticed that you use `return await ...` a lot in your code. Can you make this one similar as well for consistency's sake

server/src/auth/auth.service.ts
```diff
@@ -37,14 +95,14 @@ export class AuthService {
       subject: 'Notangles',
       text: 'Welcome to Notangles!',
       html: '<strong>Some message here!</strong>',
-    }
+    };
     sgMail
```
munjoonteo: Could convert this to `async/await` too

server/src/database/database.module.ts
```diff
@@ -1,23 +1,9 @@
 import { Module } from '@nestjs/common';
```
munjoonteo: I think you forgot to delete this

server/src/friend/dtos/friend.dto.ts
```diff
@@ -0,0 +1,18 @@
+import { Prop, Schema, SchemaFactory } from '@nestjs/mongoose';
+
+export class SingleFriendRequestDto {
```
munjoonteo: You could probably just call this `FriendRequestDto`

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+  /**
+   * Return all already added friends for a user.
+   *
+   * [Defensively checked for]:
+   *  - If the user is not found.
+   *
+   * @param req: the Request route handler param.
+   * @returns Promise to an array of users who are the users friend.
+   */
+  @Get('/:userId')
```
munjoonteo: Same here

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+   *  - Non valid users are not added as friends.
+   *
+   * @param body decorator of the form SingleFriendRequestDto. Please see
+   *        confluence documentation for the interface structure.
+   * @returns Promise to an array of users who are the users friend.
+   */
+  @Post('/')
+  @UseGuards(LoginGuard)
+  async addFriend(@Body() body: SingleFriendRequestDto): Promise<User[]> {
+    const checkValidUser = async (uId: string) =>
```
munjoonteo: Why didn't you declare this outside the function?

server/src/friend/dtos/friend.dto.ts
```diff
@@ -0,0 +1,18 @@
...
+  userId: string;
```
munjoonteo: What do you think of the keys being called `senderId` and `sendeeId` instead? `userId` is a bit ambiguous until you realise that there's a `friendId` field

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+      this.userService.checkIfUserExists(uId);
+    // Validity checker for the user Ids provided.
+    if (!checkValidUser(body.userId) || !checkValidUser(body.friendId)) {
```
munjoonteo: Could destructure `userId` and `friendId` here

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+   *        confluence documentation for the interface structure.
+   * @returns Promise to an array of users who are the users friend.
+   */
+  @Delete('/')
+  @UseGuards(LoginGuard)
+  async deleteFriend(@Body() body: SingleFriendRequestDto): Promise<User[]> {
+    const checkValidUser = async (uId: string) =>
+      this.userService.checkIfUserExists(uId);
+    // Validity checker for the user Ids provided.
+    if (!checkValidUser(body.userId) || !checkValidUser(body.friendId)) {
```
munjoonteo: Destructure here too

server/src/auth/auth.service.ts
```diff
@@ -3,30 +3,88 @@ import { InjectModel } from '@nestjs/mongoose';
...
         firstname: userInfo.given_name,
         lastname: userInfo.family_name,
         email: userInfo.email,
-        createdAt: new Date().toISOString().slice(0, 10),
+        createdAt: new Date(),
+        lastLogin: new Date(),
+        profileURL: userInfo.picture,
+        loggedIn: true,
+        settings: new UserSettingsDto(
+          false,
```
munjoonteo, line 31: Dark mode should be true by default ................

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+  @Post('/request')
+  @UseGuards(LoginGuard)
+  async sendFriendRequest(
+    @Body() body: SingleFriendRequestDto,
+  ): Promise<User[]> {
+    // Check for valid users.
+    const checkValidUser = async (uId: string) =>
+      this.userService.checkIfUserExists(uId);
+    // Validity checker for the user Ids provided.
+    if (!checkValidUser(body.userId) || !checkValidUser(body.friendId)) {
```
munjoonteo: And here

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+  @Delete('/request')
+  @UseGuards(LoginGuard)
+  async deleteFriendRequest(
+    @Body() body: SingleFriendRequestDto,
+  ): Promise<User[]> {
+    // Check for valid users.
+    const checkValidUser = async (uId: string) =>
+      this.userService.checkIfUserExists(uId);
+    // Validity checker for the user Ids provided.
+    if (!checkValidUser(body.userId) || !checkValidUser(body.friendId)) {
```
munjoonteo: And here

server/src/user/user.controller.ts
```diff
@@ -0,0 +1,136 @@
...
+  Put,
+  Request,
+  UseGuards,
+} from '@nestjs/common';
+
+import { LoginGuard } from 'src/auth/login.guard';
+
+import { Settings } from 'src/schemas/user.schema';
+import { UserSettingsDto, UserTimetablesDto } from './dtos/user.dto';
+import { User } from '@sentry/node';
```
munjoonteo: Extra import

server/src/user/user.controller.ts
```diff
@@ -0,0 +1,136 @@
...
+  @Post('/timetable/:userId')
+  @UseGuards(LoginGuard)
+  async createTimetable(
+    @Request() req,
+    @Body() body: UserTimetablesDto,
+  ): Promise<UserTimetablesDto[]> {
+    return await this.userService.createTimetable(body, req.params.userId);
+  }
+
+  @Put('/timetable/:userId/:timetableId')
```
munjoonteo: Redundant parameter `:timetableId`?

server/src/user/user.service.ts
```diff
@@ -0,0 +1,187 @@
...
+   * Edit a given timetable.
+   * @param userId: string of the user's google_uid.
+   * @param edittedTimetable: the timetable to be editted.
+   * @returns a Promise of the updated timetables.
+   */
+  async editTimetable(
+    userId: string,
+    edittedTimetable: UserTimetablesDto,
+  ): Promise<UserTimetablesDto[]> {
+    await this.deleteTimetable(userId, edittedTimetable.timetableId);
```
munjoonteo: You could've just used `findOneAndUpdate` here... deleting and creating is kinda lazy

server/src/user/user.service.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+   * Edit a given timetable.
+   * @param userId: string of the user's google_uid.
+   * @param edittedTimetable: the timetable to be editted.
+   * @returns a Promise of the updated timetables.
+   */
+  async editTimetable(
+    userId: string,
+    edittedTimetable: UserTimetablesDto,
```
munjoonteo: There is only one t in edited

server/src/user/user.service.ts
```diff
@@ -0,0 +1,187 @@
...
+  /**
+   * [Utility]
+   * Validity checker if a user exists or not
+   * @param userId
+   * @returns
+   */
+  async checkIfUserExists(userId: string): Promise<boolean> {
+    const user: UserDocument = await this.userModel
+      .findOne({ google_uid: userId })
+      .exec();
```
munjoonteo: Why do you use exec in some db calls but not others? Not criticising, just for my own knowledge

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+   *  - The user trying to add themselves as friends.
+   *  - Non valid users are not added as friends.
+   *
+   * @param body decorator of the form SingleFriendRequestDto. Please see
+   *        confluence documentation for the interface structure.
+   * @returns Promise to an array of users who are the users friend.
+   */
+  @Post('/')
+  @UseGuards(LoginGuard)
+  async addFriend(@Body() body: SingleFriendRequestDto): Promise<User[]> {
```
munjoonteo: Personally, I never explicitly specify the return types of functions because Typescript basically knows what it is already. I think it makes the lines too long.

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+      for (const u of getUserFriendReqs) {
+        if (u.google_uid === fId) {
+          return true;
+        }
+      }
+      return false;
+    };
+
+    /**
+     * There could be the case that the users are already
```
munjoonteo: Ok these comments are literally telling a story and they're kinda not necessary

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+      }
+      return false;
+    };
+
+    /**
+     * Performance optimisation:
+     * check before the promise all for the condition that
+     * they are already friends or not to save some time
+     */
+    const friended: boolean = await alreadyFriends(body.userId, body.friendId);
```
munjoonteo: Always destructure your objects

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+    if (!friended) {
+      let [recv, sent]: [boolean, boolean] = await Promise.all([
+        hasUserAsentReqToB(body.userId, body.friendId),
```
munjoonteo: Why would both A and B need to send friend requests to each other ? Is it not just one-way?

server/src/friend/friend.controller.ts
```diff
@@ -0,0 +1,207 @@
...
+   * [Defensively checked for]:
+   *  - The user trying to add themselves as friends.
+   *  - Non valid users are not added as friends.
+   *  - If the user has already sent a friend request, or already friends.
+   *
+   * @param body decorator of the form SingleFriendRequestDto. Please see
+   *        confluence documentation for the interface structure.
+   * @returns Promise to an array of users who are the users friend.
+   */
+  @Post('/request')
```
munjoonteo: I feel like having POST /request should only be used to **send** a friend request. Then, PUT /request should be used to **accept** a friend request

munjoonteo (changes requested):

munjoonteo:
A general comment about API design - a lot of your errors are simply ignored/sent to console. For example, `deleteFriendRequest` simply returns an empty array, and I assume a status code of 200, if a user is invalid. This is giving false information to the client. It should throw an error and thus return a 400 status code.

I also did some more reading about returning stuff after a POST/PUT/DELETE operation is successful, and it seems like the jury is pretty split on it. Some think that POST should return the object that was created. Some think that PUT should not return anything. A reminder to ask around and see what people think. The method I mentioned I was using for my thesis, namely not returning anything except 200 OK, is cleaner but has a big drawback - the client will have to re-fetch a lot of redundant information (e.g. for a POST, if only one object is created, re-fetching the previously created objects is a waste of bandwidth).

Your comments are also very extensive, which is good to see. But I believe Max has setup our backend to be self-documenting with the Swagger module provided by nest, so most of your personal comments are unnecessary. I get that this will take a while to write so I don't mind if you outsource writing documentation to me or something. It mostly involves defining the possible status codes that are returned by the endpoint and probably any query/path parameters. Body parameters are defined in the DTOs themselves.

server/src/friend/friend.controller.ts
Rayahhhmed: If A sent a friend req to B previously, and B is attempting to send one to A now (Basically accepting a friend req), they are added as friends and this data is removed from their individual collection.

server/src/database/database.module.ts
Rayahhhmed: Shall I remove the entire database folder and files associated with it then? Cause I kinda need that import for the Decorator below

munjoonteo: Oh yeah I meant you forgot to delete this file

munjoonteo (approved):

PR #587, opened 2022-10-03
Fix of the edit setting post endpoint

Rayahhhmed:
The previous endpoint defined a POST request for edit settings whereas a PUT request would be more suitable.


munjoonteo:
did u branch off your previous feature branch or feature/social timetabling why are there still 39 commits in this LOL

PR #588, opened 2022-10-03
Feature/user settings fix

Rayahhhmed:
The previous endpoint defined a POST request for edit settings whereas a PUT request would be more suitable.


Rayahhhmed:
Ok all should be good now O:

PR #608, opened 2022-11-04
retry server fetches n times before failing

martanman:
Tries to fix that seemingly 60% chance of getting "Could not connect to server" when first loading


Rayahhhmed:
Was the connection failing during dev or prod?

martanman:
happens in both

munjoonteo (approved):

PR #618, opened 2022-11-17
Event start time resets to default when end time changed bug fix

manhualu:
how did this cringe bug go through???!!


client/src/components/controls/CustomEventGeneral.tsx
```diff
@@ -106,7 +105,7 @@ const CustomEventGeneral: React.FC<CustomEventGeneralProps> = ({
           onChange={(e) => {
             if (e) setEndTime(e);
             setIsInitialEndTime(false);
-            setStartTime(initialStartTime);
+            console.log('setting end time');
```
munjoonteo: console log in prod smh

munjoonteo (changes requested):

PR #685, opened 2023-03-21
Ntgl 385 sharing custom events via link

incalescence:
* Users can generate an encoded link for a custom event to share with friends 
* Custom events can be added via a encoded link 
* Preview of the event location and colour to confirm the event details are correct before adding 

image
image

Co-authored-by: Sijin Soon <sijin8828@gmail.com>


incalescence:
TODO: pull from dev

incalescence:
TODO: Copied Alert

client/src/components/controls/CustomEvent.tsx
```diff
@@ -244,6 +291,7 @@ const CustomEvent: React.FC = () => {
               <TabList onChange={(_, newEventType) => setEventType(newEventType)}>
                 <Tab label="General" value="General" />
                 <Tab label="Tutoring" value="Tutoring" />
+                <Tab label="Add Invite" value="Add Invite" />
```
manhualu: i feel like 'Add Invite' isn't clear enough to the user that this is where they can create their event through link.

Maybe change it to something like "Via Link"? or unless you can think of something even better?

client/src/components/controls/CustomEvent.tsx
```diff
@@ -279,22 +327,28 @@ const CustomEvent: React.FC = () => {
                 setClassCode={setClassCode}
               />
             </StyledTabPanel>
+            <StyledTabPanel value="Add Invite">
+              <CustomEventInvite link={link} setLink={setLink} />
```
manhualu: i think we should figure out something that will handle an incorrect link better

Right now, it just crashes the entire page. Instead, have a red alert appear saying the link is not valid. 

If the link is incorrect, could you also make sure that the preview of the event does not appear?

client/src/components/controls/CustomEventInvite.tsx
```diff
@@ -0,0 +1,67 @@
...
+          onChange={(e) => checkRender(e.target.value)}
+          variant="outlined"
+          fullWidth
+          required
+          defaultValue={link}
+        />
+        <br />
+      </StyledListItem>
+      {eventPreview && (
+        <PreviewCard sx={{ bgcolor: event.color }}>
```
manhualu: can you move this to under the styled(Card) for Preview Card?

client/src/components/controls/CustomEventInvite.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <StyledListItemText>
+            <strong>{event.name}</strong>
+            <br />
+            <LocationOn sx={{ fontSize: '12px' }} />
```
manhualu: this too please :3

client/src/utils/createEvent.ts
```diff
@@ -31,3 +31,32 @@ export const createNewEvent = (
...
+      color: color,
+    },
+    time: {
+      day: day,
+      start: startTime,
+      end: endTime,
+    },
+  };
+  return newEvent;
+};
```
manhualu, line 73: this is similar to the createNewEvent function, can you try to use that function to reduce repetition?

client/src/components/controls/CustomEvent.tsx
```diff
@@ -202,18 +222,45 @@ const CustomEvent: React.FC = () => {
...
     // DroppedCards will not have the updated days and it will crash
     // (which is understandable since it's breaking React best practices by not being purely functional)
-    if (daysShort.indexOf(day) == 5) {
+    if (day == 5) {
       const MondayToSaturday: string[] = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'];
-
       setDays((prev: string[]) => (prev.length > MondayToSaturday.length ? [...prev] : MondayToSaturday));
-    } else if (daysShort.indexOf(day) == 6) {
+    } else if (day == 6) {
       setDays(['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday']);
```
Rayahhhmed: I recommend having a const array with the days and use .slice(...) instead of repeating code.

client/src/components/controls/CustomEvent.tsx
sssijin: updated to via link bc I too could not think of anything else lmao

client/src/components/controls/CustomEvent.tsx
sssijin: done!

client/src/components/controls/CustomEventInvite.tsx
sssijin: done!

client/src/components/controls/CustomEventInvite.tsx
sssijin: done!

client/src/utils/createEvent.ts
sssijin: done! Originally we created a new function because of different types. I think the new functions have less repetition but I welcome any function name suggestions, parseAndCreateNewEvent is not my best work HAHA

client/src/components/controls/CustomEvent.tsx
sssijin: done! I think it should be correct (tested creating sat and sun events).

manhualu (approved):

Rayahhhmed (approved):

PR #761, opened 2023-09-03
Fix/small css changes

manhualu:
- Footer is now responsive
- Top right of unscheduled column is now rounded

### DIFF
--- a/client/src/components/Footer.tsx
+++ b/client/src/components/Footer.tsx
@@ -10,7 +10,7 @@ const FooterContainer = styled(Box)`
   margin-bottom: 25px;
 
   & div {
-    max-width: 50%;
+    max-width: 700px;
     margin: 0 auto;
   }
 `;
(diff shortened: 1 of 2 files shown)


client/src/components/Footer.tsx
```diff
@@ -10,7 +10,8 @@ const FooterContainer = styled(Box)`
   margin-bottom: 25px;
 
   & div {
-    margin: 1vh 20vh;
+    max-width: 50%;
```
lhjt: I would suggest using a pixel value instead of % here: consider the case of on mobile where you are setting the max width to be 50% of an already super small screen.

Experiment around a bit to see what looks good but something around the ~600px mark would probably work well for what you are looking for? Then set a padding of like 1 or 2 rem so that the text doesn't reach the edges in mobile.

lhjt (changes requested):

Issue #856, opened 2024-07-03
`ctrl + backspace` is not a good keybind for deleting planner.


ollibowers:
My dearest sibling team, 

I noticed you added `ctrl + backspace` as a keybind for removing the planner. This keybind is normally used to delete whole words at a time, and is also very commonly used.  

Can you please change this to something else OR not capture when in a input box? It makes the search bar incredibly unusable for people with muscle memory. 

With warmest regards,
\- Circles Directors

Rayahhhmed:
Thank you @ollibowers  !  We will get this sorted :)

ollibowers:
im in love with you

Rayahhhmed:
can someone close this issue

mark-trann:
yeeting this into the abyss

PR #866, opened 2024-07-21
added cmd+shift+x / ctrl+shift+x as a deletion key combination

Rayahhhmed:
QOL fix with this, removing timetables are triggered by `cmd + delete` and is annoying when the timetable dialogue shows up. So this should change the hotkeys to something that is less likely to interfere with this behaviour! 
Linking the OG issue #856 !


Shzmj (approved):
nice work lads

PR #926, opened 2024-12-23
Add commit hash to client sidebar

lhvy:
This but the commit hash will be injected during Docker build instead of `dev` (which comes from my `.env` file)

image


lhvy:
Did review off GitHub - changes added, and given a rough OK. Will monitor deployment in case anything fails...

PR #930, opened 2025-01-23
Fix: Undergraduate and postgraduate courses with same code are both selected

mark-trann:
### Overview

- Currently, courses with the same code that differs in the career (`pgrd` and `ugrd`) are both rendered to be selected in the `CourseSelect` component and cannot display the correct stats for the one selected (i.e. will only display ugrd stats even when pgrd course was selected)

### Tech Debt

- Lack of frontend tests don't allow us to determine non-breaking changes. Currently will need to manually test and refocus on fix if another bug is introduced.
- Cannot differentiate between pgrd and ugrd version on the CourseSelect tags

### Fix

image

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -242,7 +242,7 @@ const App: React.FC = () => {
       prev = { ...prev };
 
       try {
-        prev[classData.id][classData.activity] = classData;
+        prev[classData.courseId][classData.activity] = classData;
       } catch (err) {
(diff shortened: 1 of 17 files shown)


lhvy:
This is going to require back end changes too.

lhvy:
Waiting on backend implementation to catch up

lhvy:
Back end should be good now - we should get this done soon.

PR #932, opened 2025-02-21
Update sponsor list and year

mark-trann:
Updating the sponsor list and year to match 2025 details.


lhvy (approved):
Looks good to me, thanks.

PR #941, opened 2025-03-10
chore(deps): bump to node v22 in CI

lhvy:


ChinoGoblino (approved):
Seems good. Only breaking changes I saw was with assert

PR #957, opened 2025-04-09
Set new classes going to unscheduled as default

lhvy:
Why? The automatically selected times are not usually people's final choice, and it's nicer to users to just drag in the times manually, particularly if they're scheduling entire courses at a time.


mark-trann (approved):
lgtm

PR #962, opened 2025-04-13
Move the term select button left of course search

### LINKED ISSUE #954: Moving the term select button from sidebar to main-page
### Description

We want to move the term selector button from the sidebar onto the same row of the "Create Event" and "Auto-timetable"

Image

### Requirements

- It's moved 👍

KailashSiva:
Resolves #954

### DIFF
--- a/client/src/components/sidebar/TermSelect.tsx
+++ b/client/src/components/sidebar/TermSelect.tsx
@@ -4,7 +4,6 @@ import {
   MenuItem,
   Select,
   SelectProps,
-  Tooltip,
   useMediaQuery,
   useTheme,
 } from '@mui/material';
@@ -23,6 +22,7 @@ const StyledInputLabel = styled(InputLabel)(({ theme }) => ({
 
 const StyledSelect = styled(Select)(({ theme }) => ({
   color: theme.palette.primary.main,
+  height: '55px',
   width: '100%',
   '& .MuiOutlinedInput-notchedOutline': {
     borderColor: theme.palette.primary.main,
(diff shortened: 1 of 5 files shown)


lhvy:
Could you please add a quick screen recording interacting with the button and expanding/closing the sidebar? Thanks

client/src/components/sidebar/TermSelect.tsx
```diff
@@ -84,7 +85,7 @@ const TermSelect: React.FC<TermSelectProps> = ({ collapsed, handleExpand }) => {
   const theme = useTheme<ThemeType>();
   const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
 
-  const [open, setOpen] = useState(false);
+  const [closed, setClosed] = useState(true);
```
lhvy: Was there a reason to change this from open to closed? Just convention to use open here, and picking closed means our `open` props need to be set to `!closed` now...

client/src/components/sidebar/TermSelect.tsx
```diff
@@ -116,45 +117,26 @@ const TermSelect: React.FC<TermSelectProps> = ({ collapsed, handleExpand }) => {
     setAssignedColors(displayTimetables[termName!][defaultStartTimetable].assignedColors);
   };
 
-  const handleMouseDown = (event: any) => {
```
lhvy, line 119: Why did we lose this?

KailashSiva: the 'clicking on term select' event did not propagate when the button was clicked, meaning the sidebar did not collapse when the term select button was clicked.

client/src/components/sidebar/TermSelect.tsx
KailashSiva: yep this is confusing. will change back sorry.

lhvy (approved):

PR #972, opened 2025-04-17
Add courseId field to class in database

lhvy:
This is in preparation to merge #930 (backend changes are required too)


server/prisma/migrations/migration_lock.toml
```diff
@@ -1,3 +1,3 @@
 # Please do not edit this file manually
 # It should be added in your version-control system (e.g., Git)
-provider = "postgresql"
\ No newline at end of file
+provider = "postgresql"
```
mark-trann, line 3: **nitpick (non-blocking)**: should end file with a newline

server/prisma/schema.prisma
```diff
@@ -71,6 +71,7 @@ model Class {
   term        String
   year        String
   courseCode  String
+  courseId    String?
```
mark-trann, line 74: **question (non-blocking)**: wouldn't we want to have the courseId non-optional?

mark-trann (approved):
lgtm

server/prisma/migrations/migration_lock.toml
lhvy: That's the change here, it does now.

server/prisma/schema.prisma
lhvy: As part of the transition, it has to start optional. (Prisma won't let it be mandatory without a default.) We'll manually fill the field in, and then transition it to mandatory, then remove the course code field when we're happy with the code.

PR #973, opened 2025-04-24
Update Prisma schema to populate course id field

lhvy:
- First update populates the course id field for classes in the user database
- Second update makes the field mandatory

Once this is merged in, we are safe to start updating the backend and frontend accordingly


mark-trann (approved):
lgtm 👍

PR #994, opened 2025-05-25
Fix authorisation for all backend routes

lhvy:
image

This PR adds authorisation to all backend routes (and disables friend and group routes since we aren't actively using them). This involves:

- Ensuring auth state is correctly held in session, and logging out redirects after clearing said session
- Cookies are sent to the BE from the FE on all API requests
- All user routes are locked behind authentication guards
- All user routes that get/modify data validate any provided user/timetable IDs

Please note that it is still _insane_ that we just send the user ID in the body/URL of these requests rather than just getting it directly from the session, but this is supposed to be a hotfix until the rewrite is done.

### DIFF
--- a/server/src/user/user.controller.ts
+++ b/server/src/user/user.controller.ts
@@ -6,16 +6,29 @@ import {
   Param,
   Post,
   Put,
+  UseGuards,
+  Request,
+  HttpException,
+  HttpStatus,
 } from '@nestjs/common';
 import { ClassDto, EventDto, InitUserDTO, TimetableDto } from './dto';
 import { UserService } from './user.service';
+import { AuthenticatedGuard } from 'src/auth/authenticated.guard';
 
 @Controller('user')
 export class UserController {
   constructor(private userService: UserService) {}
 
+  @UseGuards(AuthenticatedGuard)
   @Get('profile/:userId')
-  getUserInfo(@Param('userId') userId: string) {
+  getUserInfo(@Request() req, @Param('userId') userId: string) {
+    if (!req.user || userId !== req.user.userinfo.sub) {
+      throw new HttpException(
+        'You can only access your own profile information.',
+        HttpStatus.FORBIDDEN,
+      );
+    }
+
     return this.userService.getUserInfo(userId).then((data) => {
       return {
         status: 'Successsfully returned user profile',
@@ -24,8 +37,16 @@ export class UserController {
     });
   }
 
+  @UseGuards(AuthenticatedGuard)
   @Put('profile')
-  setUserInfo(@Body('data') data: InitUserDTO) {
+  setUserInfo(@Request() req, @Body('data') data: InitUserDTO) {
+    if (!req.user || data.userID !== req.user.userinfo.sub) {
+      throw new HttpException(
+        'You can only edit your own profile information.',
+        HttpStatus.FORBIDDEN,
+      );
+    }
+
     return this.userService.setUserProfile(data).then((res) => {
       return {
         status: 'Successfully created user!',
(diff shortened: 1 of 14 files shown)


server/src/user/user.service.ts
```diff
@@ -360,7 +378,28 @@ export class UserService {
...
+          some: {
+            userID: userID,
+          },
+        },
+      },
+    });
+
+    if (!timetable) {
+      throw new HttpException(
+        'Not allowed to delete this timetable',
```
mark-trann, line 398: nitpick: would this be a case of the timetable not existing instead of user not being allowed to delete it?

server/src/user/user.controller.ts
```diff
@@ -104,18 +169,32 @@ export class UserController {
...
-      return {
-        status: 'Successfully deleted timetable',
-        data: { timetableId: id },
-      };
-    });
+  deleteUserTimetable(
+    @Request() req,
+    @Param('timetableId') timetableId: string,
+  ) {
+    return this.userService
```
mark-trann, line 178: Now that the underlying function has been implemented to use userId, would we want to have the `userId !== req.user.userinfo.sub` check here?

server/src/user/user.controller.ts
```diff
@@ -91,11 +146,21 @@ export class UserController {
...
 
+  @UseGuards(AuthenticatedGuard)
   @Put('timetable')
   async editUserTimetable(
+    @Request() req,
     @Body('userId') userId: string,
     @Body('timetable') timetable: TimetableDto,
   ) {
+    if (!req.user || userId !== req.user.userinfo.sub) {
+      // This is not really doing anything if the underlying function doesn't make use of the provided userId.
```
mark-trann, line 157: Remove comment if userId is now relevant with added implementation.

server/src/user/user.service.ts
```diff
@@ -253,6 +253,24 @@ export class UserService {
...
+          some: {
+            userID: _userID,
+          },
+        },
+      },
+    });
+
+    if (!timetable) {
+      throw new HttpException(
+        'Not allowed to edit this timetable',
```
mark-trann, line 269: same nitpick/question here

server/src/user/user.service.ts
lhvy: It's bad practice to leak this information. Better to have a generic error message so that you can't work out if it exists or not if you don't own it.

server/src/user/user.service.ts
lhvy: Same as earlier.

server/src/user/user.controller.ts
lhvy: No `userId` passed in to the request. This one is actually safe with the underlying function fixed.

server/src/user/user.controller.ts
lhvy: Maybe I could change how the comment is worded, but I wanted to leave it here to make it clear that this check needs to be in conjunction with a safety check in the underlying function.

mark-trann:
# Test Results

### Theme

> before

image

> after (modified -> logged out -> logged back in

image

✅ 

### Timetabling

> before reschedule

image

> after reschedule

image

✅ 

> before timetable delete

image

> after timetable delete

image

✅ 

> before timetable rename

image

> after timetable rename

image

✅ 

### API

> getting timetable before logout

image

> getting timetable after logout

image

✅ 

### Custom Events

> before adding event

image

> after adding event

image

✅ 

> before modifying event

image

> after modifying event

image

✅ 

> before removing event

image

> after removing event

image

✅ 

### Tutoring

> before adding tutoring session

image

> after adding tutoring session

image

✅ 

> before removing tutoring session

image

> after removing tutoring session

image

✅ 

### Multiple Terms

> before updating t1 & t2

image

image

> after updating t1 & t2

image

image

✅

mark-trann (approved):
testing results show correct auth implementation - lgtm!

PR #1013, opened 2025-06-19
Add backend routes for courses and classes

### LINKED ISSUE #1000: Update Backend Route for Courses and Classes
## What is expected
- route for adding a new course
1. Route: `user/addCourse`
2. Params: req.user.id | req.body.courseId |

- route for removing a course
- updating the course assigned colour
- select a class for a course

## What to be in mind
- validate the information given from the request
- handling the graphql errors

jason4193:
### Description 🧾

- Added Route for courses:
1. add course
2. remove course
3. set course colour

- Still working on routes:
1. select classes

### Testing

By now, we testing our code by reading. Run them later for actual test.

### Checklist

- [ ] 📍 You have assigned yourself to this pull request.
- [ ] 🔗 You have linked an issue to which this pull request closes.
- [ ] 💭 Leave any relevant specific directions to reviewers.
- [ ] 👀 No secrets in clear text in the pull request.
- [ ] 🎟️ To categorise release notes, the pull request should be labelled with at least one of these labels:
  - `feature`: for application feature improvement or new features
  - `bug`: fixing something that previously wasn't working
  - `dev`: development-side related changes
  - `docker`: updates to the Docker code
  - `setup`: changes to the setup/infrastructure of the codebase
  - `test`: updates to internal testing


server/src/graphql/queries.ts
```diff
@@ -11,3 +11,26 @@ export const COURSE_EXISTS = gql(`
     }
   }
 `);
+
+export const CLASS_DETAILS = gql(`
```
lhvy, line 15: Can we slim this down to just what we need? It is ok to make multiple different GQL queries if there are clear times when we only want one or two fields compared to all of them. This will only be seen by the backend.

server/src/graphql/types.ts
```diff
@@ -0,0 +1,20 @@
+export class ClassDetails {
+  course_id: string;
+  activity: string;
+  career?: string | null | undefined;
```
lhvy: Can we try to simplify this to just `career?: string`? Unsure how simple this change will be - if it gets ridiculous to change it, don't bother. (May end up not even needing this field anyway...)

server/src/user/types.ts
```diff
@@ -15,3 +15,43 @@ export class UserSettings {
...
+}
+export class AddCourseDto extends CourseDto {
+  term: string;
+  colour: string;
+}
+export class SetCourseColourDto extends CourseDto {
+  colour: string;
+}
+
+export class ClassDetails {
```
lhvy: Probably can get rid of this and just import straight from the GraphQL types. You can do `import type` instead of a regular import, shouldn't affect performance.

server/src/user/user.service.ts
```diff
@@ -64,4 +75,227 @@ export class UserService {
       },
     });
   }
+
+  async isTimetableExists(
```
lhvy: Grammar?

server/src/user/user.service.ts
```diff
@@ -64,4 +75,227 @@ export class UserService {
...
+    userId: string,
+    timetableId: string,
+  ): Promise<boolean> {
+    const timetable = await this.prisma.timetable.findUnique({
+      where: {
+        id: timetableId,
+        userId: userId,
+      },
+    });
+    return timetable ? true : false;
```
lhvy: This could just be `return timetable`. Same for future instances of ternary that just return `true` and `false`.

server/src/user/user.service.ts
```diff
@@ -64,4 +75,227 @@ export class UserService {
...
+    return course ? true : false;
+  }
+
+  isColourCodeValid(colour: string): boolean {
+    const hexCodeRegex = /^#([0-9A-F]{3}|[0-9A-F]{6})$/i;
+    const defaultColoursRegex = /^default-[1-8]$/;
+    return hexCodeRegex.test(colour) || defaultColoursRegex.test(colour);
+  }
+
+  async getCourseIDs(timetableId: string): Promise<string[]> {
```
lhvy: `Ids` rather than `IDs`, just for consistency.

server/src/user/user.service.ts
```diff
@@ -64,4 +75,227 @@ export class UserService {
...
+    courseId: string,
+    timetableId: string,
+  ): Promise<CourseDetails> {
+    const course = await this.prisma.course.findFirst({
+      where: {
+        courseId: courseId,
+        timetableId: timetableId,
+      },
+    });
+    if (course === null) {
```
lhvy: I was in the process of confirming that `null` is exactly correct here (probably worth double checking) **but** regardless, I'd suggest replacing this with a `findFirstOrThrow`. You could then just try-catch that, or I think there is actually a way to map thrown errors into different ones?

server/src/user/user.service.ts
```diff
@@ -64,4 +75,227 @@ export class UserService {
...
+      },
+      data: {
+        colour: setCourseColourDto.colour,
+      },
+    });
+  }
+
+  async getClasses(courseId: string, timetableId: string): Promise<string[]> {
+    const course = await this.getCourse(courseId, timetableId);
+    return course.selectedClasses;
```
lhvy, line 254: No need to store in variable, can just directly `return` the result here.

server/src/user/user.service.ts
```diff
@@ -64,4 +75,227 @@ export class UserService {
...
+  }
+
+  async differentTimeSlotsExist(
+    timetableId: string,
+    courseId: string,
+    classId: string,
+  ): Promise<string | null> {
```
lhvy: Optional string (`undefined`) instead of `null`?

lhvy (approved):
Looks pretty good, minor feedback comments left. Would also suggest looking here and seeing if any of these status codes could be used for when your routes succeed: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#successful_responses

server/src/user/user.service.ts
jason4193: Resolved [6b73e1](https://github.com/devsoc-unsw/notangles/pull/1013/files#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR191:~:text=%3Cstring%20%7C-,undefined,-%3E%20%7B)

server/src/user/user.service.ts
jason4193: Resolved [c29303](https://github.com/devsoc-unsw/notangles/pull/1013/commits/c293037835efe59c1447a11ed6cd9c2a3c41ef5c#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR175)

server/src/user/user.service.ts
jason4193: Resolved [c29303](https://github.com/devsoc-unsw/notangles/pull/1013/commits/c293037835efe59c1447a11ed6cd9c2a3c41ef5c#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR127-R132)

server/src/user/user.service.ts
jason4193: Resolved [c29303](https://github.com/devsoc-unsw/notangles/pull/1013/commits/c293037835efe59c1447a11ed6cd9c2a3c41ef5c#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR111)

server/src/user/user.service.ts
jason4193: Resolved [c29303](https://github.com/devsoc-unsw/notangles/pull/1013/commits/c293037835efe59c1447a11ed6cd9c2a3c41ef5c#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR79)

server/src/graphql/queries.ts
jason4193: Resolved [4a8538](https://github.com/devsoc-unsw/notangles/pull/1013/commits/4a85380cc0267ca616eb2ea13ab78f36f6151053#diff-55ad6cd15b4c27155fdaf31f4ae64e5bf68df89b34ea384b39eca3a14621710aR17-R20)

server/src/graphql/types.ts
jason4193: Resolved [4a8538](https://github.com/devsoc-unsw/notangles/pull/1013/commits/4a85380cc0267ca616eb2ea13ab78f36f6151053#diff-14b7a03548f938dfbcc9d1c118d6caf17b2732a53fc855f2ec699ca598959614R1-R4)

server/src/user/types.ts
jason4193: Resolved [4a8538](https://github.com/devsoc-unsw/notangles/pull/1013/commits/4a85380cc0267ca616eb2ea13ab78f36f6151053#diff-24249b2b9b87fbbfd3b84e8818ed959c17a5709562c12c0530178981da7b1017L31-L49)

server/src/user/user.controller.ts
```diff
@@ -50,4 +61,88 @@ export class UserController {
...
+  @Get('courses/:timetableId')
+  @UseGuards(AuthenticatedGuard)
+  async getCourseIds(
+    @Req() req: AuthenticatedRequest,
+    @Param('timetableId') timetableId: string,
+  ) {
+    return await this.userService.getCourseIds(req.user.id, timetableId);
+  }
+
+  @Post('course')
```
lhvy: Any reason the post route takes the timetable ID and course ID in the request body but the delete route takes it as URL parameters? Is this HTTP best practice, or just an inconsistency we should fix?

lhvy: IMO, path parameters should be used for resource hierarchy, i.e. /course -> timetable ID -> course ID. And then any specific _data_ like colours, classes etc should go in the body. (Then optional things like limits, sorting orders, stuff we aren't currently using AFAIK would be query parameters.)

server/src/user/user.service.ts
```diff
@@ -64,4 +76,327 @@ export class UserService {
...
+          id: true,
+          selectedClasses: true,
+        },
+        where: {
+          courseId: courseId,
+          timetableId: timetableId,
+        },
+      });
+    } catch (error) {
+      if (error instanceof HttpException) {
```
lhvy: How would this error be an instance of an `HttpException`? Isn't it coming from Prisma...

server/src/user/user.service.ts
```diff
@@ -64,4 +76,327 @@ export class UserService {
...
+        'Multiple different time slots found for the same activity',
+        HttpStatus.CONFLICT,
+      );
+      if (differentTimeSlotsExist.length === 1 && differentTimeSlotsExist[0]) {
+        return differentTimeSlotsExist[0].class_id;
+      } else {
+        return undefined;
+      }
+    } catch (error) {
+      if (error instanceof HttpException) {
```
lhvy, line 287: Again - is it possible for us to actually receive a `HttpException` here? (Note: I haven't checked for this one.)

server/src/user/user.service.ts
```diff
@@ -64,4 +76,327 @@ export class UserService {
...
+    classId: string,
+  ): Promise<void> {
+    const classDetails = await this.graphqlService.getClassDetails(classId);
+    validate(
+      classDetails !== undefined,
+      'Class does not exist',
+      HttpStatus.NOT_FOUND,
+    );
+    // Assertion to help TypeScript understand that classDetails is defined
+    if (!classDetails) {
```
lhvy: Would be faster at runtime to just use `classDetails!` instead of `classDetails?`

server/src/user/user.service.ts
```diff
@@ -64,4 +76,327 @@ export class UserService {
...
+      classDetails,
+      course.selectedClasses,
+    );
+    if (differentTimeSlotClassId) {
+      await this.removeClassFromCourse(course, differentTimeSlotClassId);
+    }
+    await this.addSelectedClass(course, classId);
+  }
+
+  async addSelectedClass(
```
lhvy: Is it possible to make these very powerful methods private? Just the ones that do 0 validation and could be easily connected directly to a public API route by accident? Just now I wrote a very long feedback comment about this function without realising that it was only being called somewhere that **did** have sufficient validation.

lhvy (approved):
Looking good, really happy with the refactor. Just a few minor comments now.

server/src/user/user.service.ts
jason4193: Try to ensure every route only return httpExpection. If there are error for any internal issues like  Prisma, Graphql, other BE error, simply return 500 Server Internal Issues. Thoughts on that?

server/src/user/user.controller.ts
jason4193: Resolved [984c7e](https://github.com/devsoc-unsw/notangles/pull/1013/commits/984c7e1c2832fa3af4f3d9d1fd8121c87e7b2de6#diff-fa60de16eb8305db8ad69ad1ef526d1eb50ca04b8c166b9de5d5bdab726ee117R73)

server/src/user/user.service.ts
jason4193: Resolved [984c7e](https://github.com/devsoc-unsw/notangles/pull/1013/commits/984c7e1c2832fa3af4f3d9d1fd8121c87e7b2de6#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR141)

server/src/user/user.service.ts
jason4193: remain check if error is`HttpException` as the `validate()` inside try will throw different `HttpException` [validation function](https://github.com/devsoc-unsw/notangles/pull/1013/files#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR273-R276)

server/src/user/user.service.ts
jason4193: Resolved [984c7e](https://github.com/devsoc-unsw/notangles/pull/1013/commits/984c7e1c2832fa3af4f3d9d1fd8121c87e7b2de6#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR307-R325)

server/src/user/user.service.ts
jason4193: Resolved [984c7e](https://github.com/devsoc-unsw/notangles/pull/1013/commits/984c7e1c2832fa3af4f3d9d1fd8121c87e7b2de6#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR334-R350)

server/src/user/user.controller.ts
```diff
@@ -49,4 +60,102 @@ export class UserController {
...
+      req.user.id,
+      timetableId,
+      courseId,
+      colour,
+    );
+  }
+
+  @Get('classes/:timetableId/:courseId')
+  @UseGuards(AuthenticatedGuard)
+  async getSelectedClassesId(
```
lhvy: I think the plural of `classId` would be `classIds` rather than `classesId`. Wondering though if this should just be called `getSelectedClasses` for consistency? None of the other functions mention `id` in the names explicitly, rather the id is implied in parameter name or return type information. Thoughts?

server/src/user/user.service.ts
```diff
@@ -64,4 +70,324 @@ export class UserService {
...
   }
+
+  async isTimetableOwnedByUser(
+    userId: string,
+    timetableId: string,
+  ): Promise<boolean> {
+    const timetable = await this.prisma.timetable.findUnique({
+      where: {
+        id: timetableId,
+      },
```
lhvy, line 84: Can we add a select here to make sure we only grab the user ID? Should speed things up I think.

lhvy (approved):

server/src/user/user.controller.ts
jason4193: Resolved [cbc8ec](https://github.com/devsoc-unsw/notangles/pull/1013/commits/cbc8ec23ae4fef071eb66035619c08904ebae364#diff-fa60de16eb8305db8ad69ad1ef526d1eb50ca04b8c166b9de5d5bdab726ee117R118)

server/src/user/user.service.ts
jason4193: Resolved [cbc8e](https://github.com/devsoc-unsw/notangles/pull/1013/commits/cbc8ec23ae4fef071eb66035619c08904ebae364#diff-1b3ed7c05b4819dcf7a6c7ed2909b5e0091cd73fb1a41b1760e8c38ddb7a5cedR79-R81)

server/src/user/user.controller.ts
jason4193: Resolve [20fea](https://github.com/devsoc-unsw/notangles/pull/1013/commits/20fea88bf4507c8681c153189d229a1dff8cb38c#diff-fa60de16eb8305db8ad69ad1ef526d1eb50ca04b8c166b9de5d5bdab726ee117R118)

PR #1018, opened 2025-07-03
Update backend routes for events

fox-58:
### Description 🧾

_Context of PR, summary of changes, as well as any relevant technical decisions/motivations._

### Testing

_Replace this line with instructions on how to test your changes, as well as any relevant images for UI changes._

### Checklist

- [x] 📍 You have assigned yourself to this pull request.
- [ ] 🔗 You have linked an issue to which this pull request closes.
- [ ] 💭 Leave any relevant specific directions to reviewers.
- [x] 👀 No secrets in clear text in the pull request.
- [ ] 🎟️ To categorise release notes, the pull request should be labelled with at least one of these labels:
  - `feature`: for application feature improvement or new features
  - `bug`: fixing something that previously wasn't working
  - `dev`: development-side related changes
  - `docker`: updates to the Docker code
  - `setup`: changes to the setup/infrastructure of the codebase
  - `test`: updates to internal testing


server/src/user/user.controller.ts
```diff
@@ -49,4 +49,73 @@ export class UserController {
...
+
+  @Get('event/:timetableId/:eventId')
+  @UseGuards(AuthenticatedGuard)
+  async getEventById(
+    @Req() req: AuthenticatedRequest,
+    @Param('eventId') eventId: string,
+    @Param('timetableId') timetableId: string
+  ) {
+      try {
+        const eventDetails = await this.userService.getEvent(eventId, timetableId);
```
lhvy: Can just directly return the awaited value here, no need for variable.

server/src/user/user.service.ts
```diff
@@ -64,4 +66,79 @@ export class UserService {
       },
     });
   }
+
+  async getEvent(eventId: string, timetableId: string): Promise<EventParameters | null> {
```
lhvy: What cases does this promise actually return null? If this is explicitly for the case when you throw an error, that isn't _really_ the way to indicate that. (I'd just remove the null in that case.)

server/src/user/user.service.ts
```diff
@@ -64,4 +66,79 @@ export class UserService {
...
+      throw new HttpException('Event not found in the specified timetable', HttpStatus.NOT_FOUND);
+    }
+    return course;
+  }
+
+  async addEvent(eventParameters: EventParameters): Promise<void> {
+    await this.prisma.event.create({
+      data: {
+        id: eventParameters.id,
+        colour: eventParameters.colour,
```
lhvy: We will need to change this to use @jason4193's colour validation function once it's merged.

server/src/user/user.service.ts
```diff
@@ -64,4 +66,79 @@ export class UserService {
...
+        dayOfWeek: eventParameters.dayOfWeek,
+        start: eventParameters.start,
```
lhvy: I wonder if we need some validation on these... Might be worth checking FE code/discussing what we think makes sense to represent event times.

server/src/user/user.service.ts
```diff
@@ -64,4 +66,79 @@ export class UserService {
...
+        timetable: {
+          connect: {
+            id: eventParameters.timetableId
+          }
+        }
+      }
+    })
+  }
+
+  async removeEvent(eventId: string, timetableId: string): Promise<void> {
```
lhvy: Wondering if it makes sense to omit the `timetableId` field on all of these routes. The `eventId` is already unique across all users, across all timetables. Thoughts?

lhvy (changes requested):

server/src/user/types.ts
```diff
@@ -15,3 +17,13 @@ export class UserSettings {
...
+
+export class EventParameters {
+  id: string;
+  colour: string;
+  dayOfWeek: number;
+  start: number;
+  end: number;
+  type: EventType;
+  timetableId: string;
+}
```
jason4193: Might need to obtain everything about the event from prisma. Since only prisma db contains the details of the event that being used on FE.

server/src/user/user.controller.ts
```diff
@@ -49,4 +49,73 @@ export class UserController {
     await this.userService.setSettings(req.user.id, settings);
     return;
   }
+
+  @Get('event/:timetableId/:eventId')
+  @UseGuards(AuthenticatedGuard)
+  async getEventById(
```
jason4193: Need an additional function for getting all events'ID by using timetableID. Right now the FE is not holding any information about the Event.

jason4193: Alternative solution could be return EventID to FE after creating a new event. Then FE storing it as id attribute?

Thoughts on which way better?

server/src/user/user.service.ts
jason4193: Might be a good idea to hold the validation after combining #1000 into server-rewrite branch?

server/src/user/user.service.ts
jason4193: Agree on using eventID directly.

server/src/user/types.ts
fox-58: You mean importing from server/src/generated/prisma/models/Event.ts?

jason4193: Since the `/generated/` dir is auto generated from the schema. Not sure can we simply import it from that file. What I am thinking is having a duplicate type in here but with all most of the detail in our schema as all the information of the event will be displayed on FE.

server/src/user/user.controller.ts
lhvy: I think the required event information should come from the get timetable route. This may already be covered in #1023. Once the frontend is updated though, it may be beneficial from a performance viewpoint to add a route that gets all events from one timetable... 🤔 (Problem for later)

server/src/user/user.controller.ts
```diff
@@ -1,4 +1,4 @@
-import { Body, Controller, Get, Post, Req, UseGuards } from '@nestjs/common';
```
lhvy, line 1: Do we need these imports?

server/src/timetable/types.ts
```diff
@@ -14,3 +16,12 @@ export class UserTimetable {
   term: string;
   primary: boolean;
 }
+
+export class EventParameters {
+  id: string;
+  colour: string;
+  dayOfWeek: number;
+  start: number;
```
lhvy: Can we get some comments as to what unit this number is? Also in the Prisma schema file thanks.

server/src/graphql/queries.ts
```diff
@@ -20,3 +20,12 @@ export const CLASS_DETAILS = gql(`
     }
   }
 `);
+
+export const EVENT_DETAILS = gql(`
```
lhvy: Is this query (or its types, or the function in the event service) actually used anywhere? Events are user created - I don't think they'll appear in the GraphQL API (Unless this exists just for tutoring? But I think that is client side...)

server/src/timetable/timetable.service.ts
```diff
@@ -436,4 +461,75 @@ export class TimetableService {
       }),
     ]);
   }
+
+  async addEvent(
```
lhvy, line 493: How are these functions called? Do we have API routes for this set up in the controller?

server/src/timetable/timetable.service.ts
```diff
@@ -71,6 +72,30 @@ export class TimetableService {
...
+    userId: string,
+    eventId: string
+  ): Promise<EventParameters> {
+    try {
+      const event = await this.prisma.event.findFirstOrThrow({
+        select: { id: true, colour: true, dayOfWeek: true, start: true, end: true, type: true , timetable: { select: { userId: true } }},
+        where: { id: eventId  }
+      });
+
+      if (!event || event.timetable.userId !== userId) {
```
lhvy: The first check here won't fail. You are calling `findFirstOrThrow`, so it will throw an error if it fails. Have a look at some of the timetable routes for how they handle this.

server/src/timetable/timetable.service.ts
```diff
@@ -71,6 +72,30 @@ export class TimetableService {
...
+        throw new HttpException('Event not found or access denied', HttpStatus.NOT_FOUND);
+      }
+
+      const { timetable, ...eventData } = event;
```
lhvy: I think you could get the event data directly here if you just didn't select the timetable earlier.

server/src/timetable/timetable.service.ts
```diff
@@ -436,4 +461,75 @@ export class TimetableService {
...
+    const timetableExists = await this.isTimetableOwnedByUser(
+      userId,
+      timetableId,
+    );
+    validate(timetableExists, 'Timetable does not exist', HttpStatus.NOT_FOUND);
+
+    const colourValid = this.isColourCodeValid(eventDetails.colour);
+    validate(colourValid, 'Colour code is not valid', HttpStatus.BAD_REQUEST);
+
+    if (!(eventDetails.type in EventType)) {
```
lhvy, line 507: Curious if this is actually possible. Might be worth testing to see what happens when you pass an invalid type... Ask @jason4193 about using Postman later.

server/src/timetable/timetable.service.ts
```diff
@@ -436,4 +461,75 @@ export class TimetableService {
...
+      throw new HttpException('Invalid event type', HttpStatus.BAD_REQUEST);
+    }
+
+    const eventType: EventType = EventType[eventDetails.type as keyof typeof EventType];
```
lhvy: What does this line actually do? Isn't `eventDetails.type` of the correct type already?

server/src/timetable/timetable.service.ts
```diff
@@ -436,4 +461,75 @@ export class TimetableService {
...
+        dayOfWeek: eventDetails.dayOfWeek,
+        start: eventDetails.start,
+        end: eventDetails.end,
+        type: eventType,
+        timetable: { connect: { id: timetableId } },
+      }
+    })
+  }
+
+  async removeEvent(
```
lhvy: ~~Permissions check?~~ Realised the get function does the permissions check... Wonder if it is really optimal to pull the entire event just for this though... Thoughts @jason4193 ?

server/src/timetable/timetable.service.ts
```diff
@@ -71,6 +72,30 @@ export class TimetableService {
     }
   }
 
+  async getEvent(
+    userId: string,
+    eventId: string
+  ): Promise<EventParameters> {
+    try {
+      const event = await this.prisma.event.findFirstOrThrow({
```
lhvy: Is it possible to do a find unique here? That is preferred over find first when possible.

lhvy (changes requested):

server/src/user/user.controller.ts
lhvy: Oops, meant to comment this on the addition where a bunch got added but this file is otherwise unchanged.

server/src/graphql/queries.ts
jason4193: I don't we need this GraphQL for Event, the Class would be useful as it used to compare to check if the class is already be selected before, but we don't need that for Event.

server/src/timetable/timetable.service.ts
jason4193: Agree might be more optimal by calling some validation similar to `isTimetableOwnedByUser`

jason4193 (commented):
similar to @lhvy 's comments, happy to test on those routes using postman

server/src/user/user.controller.ts
```diff
@@ -47,4 +47,6 @@ export class UserController {
   ) {
     await this.userService.setSettings(req.user.id, settings);
   }
+
+
```
mark-trann: 🪓

server/src/timetable/timetable.service.ts
```diff
@@ -436,4 +437,122 @@ export class TimetableService {
...
+    eventId: string
+  ): Promise<EventParameters> {
+    try {
+      const event = await this.prisma.event.findUnique({
+        select: { id: true, colour: true, dayOfWeek: true, start: true, end: true, type: true , timetable: { select: { userId: true } }},
+        where: { id: eventId  }
+      });
+
+      if (!event || event.timetable.userId !== userId) {
+        throw new HttpException('Event not found or access denied', HttpStatus.NOT_FOUND);
```
mark-trann: nitpick: shouldn't describe the error as "access denied" to hide details from consumer

server/src/timetable/timetable.service.ts
```diff
@@ -436,4 +437,122 @@ export class TimetableService {
...
     ]);
   }
+
+  async getEvent(
+    userId: string,
+    eventId: string
+  ): Promise<EventParameters> {
+    try {
+      const event = await this.prisma.event.findUnique({
+        select: { id: true, colour: true, dayOfWeek: true, start: true, end: true, type: true , timetable: { select: { userId: true } }},
```
mark-trann: spread this over multiple lines for readability

server/src/graphql/types.ts
```diff
@@ -2,3 +2,8 @@ export class ClassDetails {
   activity: string;
   section: string;
 }
+
+export class EventDetails {
+  activity: string;
+  section: string;
+}
```
mark-trann: would it be better to have the class extend off `ClassDetails`? It may get messy if ClassDetails change though...

server/prisma/schema.prisma
```diff
@@ -107,11 +107,15 @@ model Event {
...
-  dayOfWeek Int
-  start     Int
-  end       Int
-  type      EventType
+  title       String
+  description String?
+  location    String?
+
+  colour      String
+  dayOfWeek   Int         // 0 = Sunday, 6 = Saturday
```
lhvy: Why did we start at Sunday? Is there a library reason for this - might be unintuitive given that UI is Monday to Sunday

server/src/auth/auth.controller.ts
```diff
@@ -60,7 +60,7 @@ export class AuthController {
       userId = (res.req as AuthenticatedRequest).user.id;
     }
 
-    res.clearCookie('connect.sid');
+    // res.clearCookie('connect.sid');
```
lhvy: ?

server/src/timetable/types.ts
```diff
@@ -21,3 +23,15 @@ export enum Term {
   T2 = 'T2',
   T3 = 'T3',
 }
+
+export class EventParameters {
```
lhvy: Looks like we have this type written twice. Let's ensure we only use one consistent copy. Also, is this a DTO like the other types around it?

If this is a type being used for data coming in or out of a controller, we should keep it, but we should not be redefining types from Prisma for service/internal functions, Prisma already gives us neat types (see how we import the `User` and `Timetable` types from Prisma)

server/src/timetable/timetable.controller.ts
```diff
@@ -202,4 +203,105 @@ export class TimetableController {
...
+        HttpStatus.INTERNAL_SERVER_ERROR,
+      );
+    }
+  }
+
+  @Post('event')
+  @UseGuards(AuthenticatedGuard)
+  async addEvent(
+    @Req() req: AuthenticatedRequest,
+    @Body() body: { timetableId: string; event: EventParameters },
```
lhvy: Let's make this route behave like the other timetable routes and have the timetable ID be part of the URL (see the add course route)

server/src/timetable/timetable.controller.ts
```diff
@@ -202,4 +203,105 @@ export class TimetableController {
...
+        const dateString = Date.now().toString(36);
+        const randomness = Math.random().toString(36).substring(2);
+        return dateString + randomness;
+      };
+
+      console.log(generateUniqueId());
+
+      await this.timetableService.addEvent(
+        req.user.id,
+        generateUniqueId(),
```
lhvy: Let Prisma make the IDs for you. If you don't specify the ID, Prisma will generate a new UUID.

server/src/timetable/timetable.controller.ts
```diff
@@ -202,4 +203,105 @@ export class TimetableController {
...
+
+  @Delete('event/:eventId')
+  @UseGuards(AuthenticatedGuard)
+  async deleteEvent(
+    @Req() req: AuthenticatedRequest,
+    @Param('eventId') eventId: string,
+  ) {
+    try {
+      await this.timetableService.removeEvent(req.user.id, eventId);
+      return { message: 'Event deleted successfuly' };
```
lhvy: This message doesn't provide us with any useful information. It is easier (and cheaper) to just return 200 OK with no message.

server/src/timetable/timetable.service.ts
```diff
@@ -436,4 +442,147 @@ export class TimetableService {
...
+      where: {
+        id: eventId,
+      },
+    });
+  }
+
+  async updateEvent(
+    userId: string,
+    eventId: string,
+    eventDetails: Partial<EventParameters>,
```
lhvy: Let's update this partial type to explicitly exclude ID and timetable ID. Then we can pass just what we need (event ID) as param, which you already do.

lhvy: Need to test this exclusion at a controller level (ensuring that if you make a request including either of those fields, it gets rejected).

server/src/timetable/timetable.service.ts
```diff
@@ -436,4 +442,147 @@ export class TimetableService {
...
+  ): Promise<void> {
+    const event = await this.getEvent(userId, eventId);
+
+    await this.prisma.event.update({
```
lhvy, line 555: You can simplify the provided data first with an `Object.fromEntries` that filters out the undefined fields, and then simply call

```ts
  await this.prisma.event.update({
    where: { id: eventId },
    data: filteredData,
  });
```

server/src/timetable/timetable.service.ts
```diff
@@ -436,4 +442,147 @@ export class TimetableService {
...
+        }),
+        ...(eventDetails.start !== undefined && { start: eventDetails.start }),
+        ...(eventDetails.end !== undefined && { end: eventDetails.end }),
+        ...(eventDetails.type && { type: eventDetails.type }),
+      },
+    });
+  }
+
+  async isEventOwnedByUser(userId: string, eventId: string): Promise<boolean> {
+    const event = await this.prisma.event.findUnique({
```
lhvy: Something like this would be more optimised (only one query, find first to early exit on first (and only) result).

```ts
const hit = await this.prisma.event.findFirst({
  where: { id: eventId, timetable: { userId } },
  select: { id: true },
});
return !!hit;
```

lhvy (changes requested):

server/src/timetable/timetable.service.ts
jason4193: @lhvy Tested on the Postman, only `default-<1 to 8>`, `#<3 letters 0-9/A-F>` & `#<6 letters 0-9/A-F>` will pass the validation.
Something like `#default`, `default-9`, `#xxx` will fail

lhvy (approved):

PR #1022, opened 2025-07-03
Fix primary timetables localstorage bug by checking for version 2.0 on reload

rfquin:
### Description 🧾
TBA (I'll add later)

_Context of PR, summary of changes, as well as any relevant technical decisions/motivations._

### Testing

_Replace this line with instructions on how to test your changes, as well as any relevant images for UI changes._

### DIFF
--- a/client/src/utils/storage.ts
+++ b/client/src/utils/storage.ts
@@ -1,6 +1,7 @@
 import defaults from '../constants/defaults';
 import migrateThemes from './migrations/colourTheme';
 import migratePrimaryTimetables from './migrations/primaryTimetables';
+import { createDefaultTimetable } from './timetableHelpers';
 
 const STORAGE_KEY = 'data';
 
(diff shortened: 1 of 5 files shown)


client/src/utils/storage.ts
```diff
@@ -103,11 +53,20 @@ const storage = {
...
-    if (data.version !== 1 || data.version == null) {
+    if (data.version == null) {
       let migrated = {
         ...data,
         version: 1,
-        timetables: migrateTimetables(data.timetables),
+        timetables: migrateThemes(data.timetables),
+      };
+      storage.save(migrated);
+      return migrated;
```
lhvy: We should let migrations bubble up. i.e. If a user with version 0/no version attempts to migrate, the single function call should take them from 0 to 1, then 1 to 2, and then return.

lhvy (changes requested):

client/src/utils/storage.ts
rfquin: Yep, mb :+1:

lhvy (approved):

PR #1023, opened 2025-07-09
Timetable routes

sunnyyuqingchen:
added routes for:
GET /user/timetables
GET /user/timetables/:id/view	
POST /user/timetables
PATCH /user/timetables/:id/change-primary
DELETE /user/timetables/:id


server/src/user/user.controller.ts
```diff
@@ -49,4 +60,84 @@ export class UserController {
...
+    @Query('year') year: string,
+    @Query('term') term: string,
+  ) {
+    await this.userService.deleteTimetable(
+      req.user!.id,
+      timetableId,
+      Number(year),
+      term,
+    );
+    return;
```
lhvy: No need to write return if the function just ends with no return value.

server/src/user/user.service.ts
```diff
@@ -64,4 +68,171 @@ export class UserService {
...
+    if (numTimetables <= 1) {
+      throw new ForbiddenException('Cannot delete the last timetable.');
+    }
+
+    const timetable = await this.prisma.timetable.findFirst({
+      where: { id: timetableId, userId },
+    });
+
+    if (!timetable) {
+      throw new ForbiddenException('Timetable does not belong to this user.');
```
lhvy: Can we make all instances of no timetable for userId + timetableId a 404 instead.

lhvy (changes requested):

lhvy:
Please also look over the code + review comments for #1013 and #1018 to see what you do/don't like - feel free to add more comments there, or to update your own code to match conventions with the comments already left.

PR #1027, opened 2025-07-14
Migrate from react-beautiful-dnd to @hello-pangea/dnd

lhvy:
### Description 🧾

react-beautiful-dnd has been deprecated for a while now. It is not compatible with React 18' Strict Mode, nor is it compatible with React 19. @hello-pangea/dnd seems to have forked off of RBDnD and has provided bug fixes and updates for the package. It seems to be acting as a drop-in replacement currently, and solves my current issues with react-beautiful-dnd.

I also bumped some minor dependencies and ran `pnpm upgrade` whilst I was at it. 🥰

### Testing

Have quickly tested where we use RBDnD (as far as I'm aware, just the timetable tabs), and the functionality seems the same. Always helps to get more testing from another dev though!


mark-trann (approved):
slay

mark-trann:
fck it we ball fr

PR #1029, opened 2025-07-15
Upgrade to MUI 7

lhvy:
### Description 🧾

Still a lot of work to be done. Peer checking and testing required after each step below:

- [X] Migrate from MUI X 6 to 7
- [x] Migrate from MUI 5 to 6
- [ ] ~~Migrate to Pigment CSS~~ Deferred in #1051 
- [x] Migrate from MUI 6 to 7
- [x] Migrate from MUI X 7 to 8
- [x] Upgrade date-fns

### Testing

To be done incrementally by dev + other team members, plus one large FE test at the end.


lhvy:
Frontend looks the same to me, but a second pair of eyes testing and comparing locally would be awesome as well thanks.

client/package.json
```diff
@@ -16,41 +16,42 @@
...
-    "@sentry/node": "9.38.0",
-    "@sentry/react": "9.38.0",
-    "@tanstack/react-query": "5.84.1",
-    "@uiw/react-color": "2.7.1",
+    "@mui/icons-material": "7.3.1",
+    "@mui/lab": "7.0.0-beta.16",
+    "@mui/material": "7.3.1",
+    "@mui/system": "7.3.1",
+    "@mui/x-date-pickers": "8.11.0",
+    "@sentry/browser": "10.8.0",
```
mark-trann, line 30: we should plan to remove sentry eventually...i don't think we're using it

client/src/components/sidebar/Privacy.tsx
```diff
@@ -154,8 +152,8 @@ const Privacy: React.FC = () => {
       <StyledAccordion key={index}>
         <StyledAccordionSummary
           expandIcon={<ExpandMore />}
-          aria-controls={`privacyPanel${index}-content`}
-          id={`privacyPanel${index}-header`}
+          aria-controls={`privacyPanel${String(index)}-content`}
+          id={`privacyPanel${String(index)}-header`}
```
mark-trann, line 156: was this throwing a type error? JS's backticks should be doing its magic regardless of the type `index` is

mark-trann (commented):
good stuff lgtm. will test ui before approving

client/src/components/sidebar/Privacy.tsx
lhvy: Linter moment. I'm not opposed to this tbh, but I wonder what people think? Here is an explanation of the rule: https://typescript-eslint.io/rules/restrict-template-expressions. We could switch it off just for specific types such as number or bool, but I do note that the rule page points out the default cast may not be ideal for floating points, so this rule forces you to think more explicitly about when and how you cast.

client/package.json
lhvy: I would like to get analytics working again at some point, taking to platforms about it again currently...

mark-trann: Ah yeah tru - surely we can use up some of the devsoc budget for sentry

client/src/components/sidebar/Privacy.tsx
mark-trann: I'm not against it either. Seems like it'll be helpful preventing nullish and container types to be templated into a string. I found this to be a bit odd so had an eyebrow raised. Can reconsider this lint rule later on if it becomes a hassle 👍

mark-trann (approved):

PR #1031, opened 2025-07-15
More aggressively catch corrupted local storage

lhvy:
### Description 🧾

Motivated by seeing our second user with corrupted storage, already migrated to version 2. Something is very wrong here - we should look at where new terms get created/added, as well as all of the code that syncs data _from_ the user database to local storage. In the meantime/since we are replacing that code anyway, my fairly destructive check that gets users out of binds can be run once per refresh to try and get users out of this issue.

### Testing

Not tested yet - I will have more time later this week.

### DIFF
--- a/client/src/utils/storage.ts
+++ b/client/src/utils/storage.ts
@@ -1,4 +1,3 @@
-import { useRef } from 'react';
 import defaults from '../constants/defaults';
 import migrateThemes from './migrations/colourTheme';
 import migratePrimaryTimetables from './migrations/primaryTimetables';
@@ -9,7 +8,7 @@ const STORAGE_KEY = 'data';
 // Something is not playing nice and causing local storage to corrupt across signing in/out
 // We are in the process of migrating data to the backend, so this fix should catch people
 // for now and prevent them from seeing a blank page.
-const hasRunArrayCheck = useRef(false);
+let hasRunArrayCheck = false;
 
 const storage = {
   get: (key: string): any => {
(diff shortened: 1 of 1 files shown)


client/src/utils/storage.ts
```diff
@@ -1,10 +1,16 @@
...
 import migrateThemes from './migrations/colourTheme';
 import migratePrimaryTimetables from './migrations/primaryTimetables';
 import { createDefaultTimetable } from './timetableHelpers';
 
 const STORAGE_KEY = 'data';
 
+// Something is not playing nice and causing local storage to corrupt across signing in/out
+// We are in the process of migrating data to the backend, so this fix should catch people
+// for now and prevent them from seeing a blank page.
+const hasRunArrayCheck = useRef(false);
```
jason4193: While testing, error pop up as `TypeError: can't access property "useRef", dispatcher is null`.
Using `let hasRunArrayCheck = false` should handle the case already?

client/src/utils/storage.ts
```diff
@@ -73,6 +79,8 @@ const storage = {
           migrated.timetables[term] = createDefaultTimetable('');
```
jason4193, line 78: Require to `.save` the localStorage after reset. Since there is case where `version > 2`, but the timetable format is invalid.

jason4193 (changes requested):
After running test locally, identify these errors.
1. `useRef` type error on utils
2. case of updated timetable not save on localStorage when `version > 2`

client/src/utils/storage.ts
lhvy: Should be fine, brain is coming from different code where useRef made sense. I need a break today haha.

client/src/utils/storage.ts
lhvy: Oops yes true.

PR #1058, opened 2025-09-05
Implement Google and GitHub sign in

lhvy:
### Description 🧾

Can now log in with GitHub or Google using OAuth 2.0

### Testing

Tested locally (briefly), LGTM. Data in Prisma Studio seems ok too.


client/src/components/timetable/ExpandedClassView.tsx
```diff
@@ -175,11 +175,9 @@ const ExpandedClassView: React.FC<ExpandedClassViewProps> = ({ code, classPeriod
    * @param e The HTML event triggered
    */
   const handleLocationChange = (e: SelectChangeEvent<number>) => {
-    setSelectedIndex(e.target.value as number);
```
mark-trann, line 178: lint error?

server/.env.example
```diff
@@ -0,0 +1,22 @@
+NODE_ENV="dev"
```
mark-trann, line 1: good call on an env example

server/src/auth/google.strategy.ts
```diff
@@ -0,0 +1,56 @@
...
+      id: string;
+      name: { givenName: string; familyName: string };
+      emails: { value: string; verified: boolean }[];
+    },
+  ) {
+    const user = await this.prisma.user.upsert({
+      where: {
+        authProvider_authSubject: {
+          authProvider: 'GOOGLE',
+          authSubject: profile.id,
```
mark-trann, line 34: Is the profile id the same for every login attempt from the same gmail account?

mark-trann (approved):
lgtm! this is gonna be so sick 🔥

happy to leave login and logout routes as GET apis 👍

server/src/auth/google.strategy.ts
lhvy: yeppers

client/src/components/timetable/ExpandedClassView.tsx
lhvy: yeppers

PR #1072, opened 2025-09-17
Fuzzy search implementation for friend list

mark-trann:
works 👍 

image


mark-trann:
TODO: need to fix this implementation slightly after integration with server

lhvy (approved):

PR #1082, opened 2025-10-15
Remove old account implementation

lhvy:
### Description 🧾

* Remove old account implementation

* Delete group sidebar

* Remove timetable syncing and groups

(This is so that we don't get cooked by migration with the rewrite)


mark-trann (approved):
lgtm

